PR #2, opened 2018-04-04
Initial schema creation and graph sync

fridex:


thoth/storages/graph/models.py
```diff
@@ -1,61 +1,188 @@
...
 
-        return instance
+    release = properties.Property(properties.String)
+    epoch = properties.Property(properties.String)
+    arch = properties.Property(properties.String)
+    src = properties.Property(properties.Boolean)
+    package_identifier = properties.Property(properties.String)
 
+    @classmethod
+    def construct(cls, package_info: dict):
```
fridex: note to myself: this should be named ```from_document()``` here and else where

thoth/storages/graph/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    # Assign to instance so instance has the id associated correctly for later queries.
+    vertex.id = result['id']
+
+    return result['id'], result['existed']
+
+
+async def get_or_create_edge(g: AsyncGraphTraversalSource, edge: EdgeBase) -> tuple:
+    """Create an edge if not existed before, if the given edge already exists, get its id."""
+    source_id = edge.source.id
+    target_id = edge.target.id
```
fridex: note to myself: create an alternative that would accept directly ids so we can cache vertexes on worker/syncing side. The lookup on the caching side will be still O(N) considering variable properties, but N will be much smaller compared to N in the graph database (N = number of vertexes stored).

PR #23, opened 2018-05-15
Introduce Ceph connection check

fridex:


fridex:
Related: https://github.com/thoth-station/result-api/issues/17

goern:
>/assign @goern 

@fridex if the file is not found that will not raise an exception?!

fridex:
> @fridex if the file is not found that will not raise an exception?!

The `document_exists()` method returns true/false depending on file presence. See comment in the implemented method.

PR #30, opened 2018-06-01
 Skip python packages that do not have mercator result

fridex:


goern:
@fridex ping

fridex:
> @fridex ping

I think this is related to old results we have on Ceph already.

PR #33, opened 2018-06-25
Query unsolved Runtime Environments from DB

sushmithaaredhatdev:
Fixes: #https://github.com/thoth-station/graph-refresh-job/issues/11

### DIFF
--- a/thoth/storages/graph/janusgraph.py
+++ b/thoth/storages/graph/janusgraph.py
@@ -364,6 +364,26 @@ class GraphDatabase(StorageBase):
 
         return python_package, has_version, python_package_version
 
+    def unsolved_runtime_environments(self, package_name: str, package_version: str) -> list:
+        """Get Unsolved Runtime Environment which are not connected and attached to python package."""
+        query = self.g.V() \
+            .has('__label__', 'python_package_version') \
+            .has('package_name', package_name) \
+            .has('package_version', package_version) \
+            .inE() \
+            .has('__label__', 'solved') \
+            .outV() \
+            .dedup() \
+            .aggregate('inter_result') \
+            .V() \
+            .hasLabel('label', 'ecosystem_solver') \
+            .where(without('inter_result')) \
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/janusgraph.py
```diff
@@ -364,6 +364,26 @@ def create_pypi_package_version(self, package_name: str, package_version: str, *
 
         return python_package, has_version, python_package_version
 
+    def unsolved_runtime_environments(self, package_name: str, package_version: str) -> list:
+        """Get Unsolved Runtime Environment which are not connected and attached to python package."""
```
fridex: > **U**nsolved **R**untime **E**nvironment

No need have it upper case.

> attached to python package

Actually it is "python package version", not only python package.

thoth/storages/graph/janusgraph.py
```diff
@@ -364,6 +364,26 @@ def create_pypi_package_version(self, package_name: str, package_version: str, *
...
+        query = self.g.V() \
+            .has('__label__', 'python_package_version') \
+            .has('package_name', package_name) \
+            .has('package_version', package_version) \
+            .inE() \
+            .has('__label__', 'solved') \
+            .outV() \
+            .dedup() \
+            .aggregate('inter_result') \
```
fridex: When reading this, "inter_result" does not help me what it supposed to be. For better reading, I would suggest to rename it to something like "solvers_solved_python_package_version" or something like that.

thoth/storages/graph/janusgraph.py
```diff
@@ -364,6 +364,26 @@ def create_pypi_package_version(self, package_name: str, package_version: str, *
...
+            .V() \
+            .hasLabel('label', 'ecosystem_solver') \
```
fridex: We use properties as labels - can be used inside indexes. This needs to be changed to:

```diff
+            .has('__label__', 'ecosystem_solver') \
```

PR #131, opened 2018-09-27
Specify Python index from which the package came from

fridex:
Depends-On: https://github.com/goern/janusgraph-openshift/pull/25


goern:
regate

goern:
recheck

PR #154, opened 2018-10-10
using envvar injected by OpenShift to discover janusgraph service

goern:
using envvar that are injected by OpenShift to discover janusgraph servcie host and port, this requires that a service called "janusgraph" is created

### DIFF
--- a/thoth/storages/graph/janusgraph.py
+++ b/thoth/storages/graph/janusgraph.py
@@ -85,8 +85,8 @@ def requires_connection(func):
 class GraphDatabase(StorageBase):
     """A Gremlin server adapter communicating via a web socket."""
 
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/janusgraph.py
```diff
@@ -251,7 +251,8 @@ def get_runtime_environment(self, runtime_environment_name: str,
             )
 
             if not analysis_document_id:
-                raise NotFoundError(f"No entries for runtime environment {runtime_environment_name!r} found")
+                raise NotFoundError(
+                    f"No entries for runtime environment {runtime_environment_name!r} found")
```
fridex: :disappointed:

goern: ja... merged it because its blocking me :/

fridex: ok, but fix it

PR #212, opened 2018-11-02
Introduce unparsed flag

fridex:
Depends-On: https://github.com/goern/janusgraph-openshift/pull/29
Fixes: https://github.com/thoth-station/solver/issues/128

### DIFF
--- a/thoth/storages/graph/models.py
+++ b/thoth/storages/graph/models.py
@@ -157,15 +157,15 @@ class Solved(EdgeBase):
     solver_error = Property(properties.Boolean, default=False)
     # Issues when solving the given package (e.g. the given package does not
     # exist or the given version does not exist.)
-    # This flag is an addition to solver_error, meaning if solver_unsolvable is
+    # This flag is an addition to solver_error, meaning if solver_error_unsolvable is
     # set to True, solver_error is True as well. But NOT vice versa.
     # This behaviour is to simplify queries during recommendations.
-    solver_unsolvable = Property(properties.Boolean, default=False)
+    solver_error_unsolvable = Property(properties.Boolean, default=False)
     # Issues when parsing package names (package name does not conform to PEP - e.g. spaces in name).
-    # This flag is an addition to solver_error, meaning if solver_unparsed is
+    # This flag is an addition to solver_error, meaning if solver_error_unparsable is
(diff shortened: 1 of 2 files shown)


thoth/storages/graph/models.py
```diff
@@ -150,15 +150,22 @@ class Solved(EdgeBase):
...
     # exist or the given version does not exist.)
     # This flag is an addition to solver_error, meaning if solver_unsolvable is
     # set to True, solver_error is True as well. But NOT vice versa.
     # This behaviour is to simplify queries during recommendations.
     solver_unsolvable = Property(properties.Boolean, default=False)
+    # Issues when parsing package names (package name does not conform to PEP - e.g. spaces in name).
+    # This flag is an addition to solver_error, meaning if solver_unparsed is
+    # set to True, solver_error is True as well. But NOT vice versa.
+    # This behaviour is to simplify queries during recommendations.
+    solver_unparsed = Property(properties.Boolean, default=False)
```
goern: to me it feels like it should be called 'solver_packagename_unparsable' or 'solver_error_unparsable' ?!

fridex: It can be also version that is unparsabe. However I like the latter one as it is seen from name that is is a special case of ``solver_error``. It can be also applied to unsolvable. :+1:

thoth/storages/graph/models.py
```diff
@@ -150,15 +150,22 @@ class Solved(EdgeBase):
...
+    # This flag is an addition to solver_error, meaning if solver_error_unsolvable is
     # set to True, solver_error is True as well. But NOT vice versa.
     # This behaviour is to simplify queries during recommendations.
-    solver_unsolvable = Property(properties.Boolean, default=False)
+    solver_error_unsolvable = Property(properties.Boolean, default=False)
+    # Issues when parsing package names (package name does not conform to PEP - e.g. spaces in name).
+    # This flag is an addition to solver_error, meaning if solver_error_unparsed is
+    # set to True, solver_error is True as well. But NOT vice versa.
+    # This behaviour is to simplify queries during recommendations.
+    solver_error_unparsed = Property(properties.Boolean, default=False)
```
goern: still, why "unparsed" rather than "unparsable" ?

fridex: Adjusted.

PR #217, opened 2018-11-03
Use common date utilities for creating datetime from timestamp

fridex:
Depends-On: https://github.com/thoth-station/common/pull/194


fridex:
The codacy report is not relevant here - we have these utilities in thoth common since 0.4.0 https://github.com/thoth-station/common/blob/master/CHANGELOG.md#release-040-2018-11-04t212814

PR #240, opened 2018-11-08
Fix queries to janusgraph - aggregate by document ids

fridex:


thoth/storages/graph/janusgraph.py
```diff
@@ -394,6 +394,8 @@ def get_solver_documents_count(self) -> dict:
             .has('solver_error') \
             .has('solver_error_unsolvable') \
             .has('solver_error_unparsable') \
+            .valueMap() \
+            .select('solver_document_id')
```
goern: before this to took all documents?

fridex: I've realized it deduplicates based also on source/destination node - we simply counted all edges. We just want to count based on document ids.

PR #249, opened 2018-11-08
Make sure we know that packages are resolved for transitive ones

fridex:


fridex:
Fixed in https://github.com/thoth-station/storages/pull/386

PR #260, opened 2018-11-13
Add a query to check for solved packages

fridex:


fridex:
@goern can we have this in the "Graph Database: ..." graph on grafana?

PR #291, opened 2018-11-20
Introduce buildtime environment model

fridex:


fridex:
Marking as WIP, we will need schema changes.

PR #318, opened 2018-11-22
Introduce method for registering Python package indexes

fridex:


fridex:
The pytest failure is caused by unreleased amun-client. Will merge this to continue with work.

PR #386, opened 2018-11-29
Consider index when retrieving transitive dependencies

fridex:


fridex:
recheck

thoth/storages/graph/janusgraph.py
```diff
@@ -573,20 +573,21 @@ def retrieve_dependencies(self, package_name: str, package_version: str, index:
...
-    def retrieve_transitive_dependencies_python(self, package_name: str, package_version: str, index: str) -> list:
+    def retrieve_transitive_dependencies_python(self, package_name: str, package_version: str, index_url: str) -> list:
         """Get all transitive dependencies for the given package by traversing dependency graph."""
         query = self.g.V() \
             .has('__type__', 'vertex') \
             .has('__label__', 'python_package_version') \
             .has('ecosystem', 'pypi') \
             .has('package_version', package_version) \
             .has('package_name', package_name) \
+            .has('index_url', index_url) \
```
goern: what happens to packages without index_url or empty index_url?

fridex: As index_url is required to be set, it will be omitted from the query. We expect index_url to be always set for resolution to have predictable sources (this query is used there).

PR #442, opened 2019-01-07
Update README.rst

CermakM:
Add note about issue with hanging gremlin when using aiogremlin==3.3.1


fridex:
The coala failure is not relevant, merging.

PR #497, opened 2019-02-11
Add platform specific features to the transitive query

fridex:
This way we will be able to ask for results of platform specific solver
results.


fridex:
We will need to refresh graph database data for this.

PR #520, opened 2019-02-14
Introduce advised relationship

fridex:


fridex:
We currently do not capture errors in advises. That's something we should consider to capture - so maybe introduce provenance-checker software stack type

PR #521, opened 2019-02-14
Add python version and cuda version to graph schema

fridex:


goern:
https://zuul.thoth-station.ninja/logs/21/521/cd8725c1452fb8a5a02a6c18c20d18e750e28c5f/check/thoth-coala/1e5c00e/ara-report/result/9df8d58d-45c5-4baa-876f-00a0d11550c2/

fridex:
@goern Yes, I saw it. Would you mind to adjust the coala configuration so that it accepts longer files?

> thoth/storages/graph/janusgraph.py
|    | [NORMAL] LineCountBear:
|    | This file had 1733 lines, which is 233 lines more than the maximum limit specified.
Executing section all.python...

goern:
@fridex I dont know what the best practice in the python universe is... we should keep it at that level and add an exception for `janusgraph.py`?!

fridex:
> @fridex I dont know what the best practice in the python universe is... we should keep it at that level and add an exception for `janusgraph.py`?!

It's up to project scope. We can add directive to the janusgraph.py file to suppress this, maybe it will be better.

PR #564, opened 2019-03-18
Distinguish between runtime and buildtime environment

fridex:


fridex:
Tested in test environment.

Issue #556, opened 2019-03-14
Link PythonPackageIndex with PythonPackageVersion


fridex:
We should create an edge between `PythonPackageIndex` and `PythonPackageVersion` so that we can use this connection in queries. Also, this way we can easily check that each `PythonPackageVersion` is from an index known to Thoth (there is relevant `PythonPackageIndex` vertex). If somebody tries to install/use a `PythonPackageVersion` from an index not known to Thoth, Thoth should error out (only administrator of Thoth can add new `PythonPackageIndex` vertexes).

fridex:
CC @pacospace

pacospace:
Added new edge in #574

fridex:
Implemented.

PR #574, opened 2019-03-25
New Edge between PythonPackageVersion and PythonPackageIndex #556

pacospace:


fridex:
> Check coala issues.

Ah, ok... so coala issues were not introduced by this patch.

PR #576, opened 2019-03-29
Adjust method signatures

fridex:


goern:
looks like black is fighting with coala about spaces?!

PR #585, opened 2019-04-11
Improve schema handling

fridex:
* entries in generated file are always sorted alphabetically
* give edges and vertexes name if autogenerated names are not nice


pacospace:
lgtm!

PR #592, opened 2019-04-16
Implemented get_all_python_packages_count method for Dgraph

pacospace:


thoth/storages/graph/schema.rdf
```diff
@@ -39,7 +39,7 @@ depends_on_label: string @count @upsert @index(exact) .
 depends_on: uid @reverse @count .
 os_name: string @index(exact) .
 os_version: string @index(exact) .
-solver_error: bool .
+solver_error: bool @index(bool) .
```
fridex: This one will be discarded on the next schema update (as the file is autogenerated), I will adjust the corresponding model.

pacospace: Yes thanks and maybe also others will come out, as Dgraph don't allow querying without index.

PR #599, opened 2019-04-18
Remove checks which are already present in _create_python_package_record

fridex:


fridex:
@pacospace The check is triggered if `verify_index` is set to true - its set only for solver syncs as that us creating packages for adviser resolution: https://github.com/thoth-station/storages/blob/2582bfe5a8b6a7f8a8d91f65c4a9ffcef683bd9e/thoth/storages/graph/dgraph.py#L1163

The check has basically no overhead as syncs are triggered with vertex cache - if there are multiple queries to dgraph which would resolve in same vertex, the cache is hit and matching vertex is returned (no actual network traffic/querying on dgraph side).

PR #600, opened 2019-04-18
Dgraph fix client connection for query

pacospace:


thoth/storages/graph/models_base.py
```diff
@@ -175,7 +175,7 @@ def _do_query(cls, client: DgraphClient, keys: dict) -> List["Element"]:
             }
         }
         """ % (label, filter_str, "\n".join(cls.get_properties().keys()))
-        query_result = client.query(query)
+        query_result = client.txn().query(query)
```
fridex: Hm... the defaul is `read_only=False` which is bad as transaction keeps opened. I will fix this in a separate PR.

PR #604, opened 2019-04-18
Implement query for retrieving solver error count

fridex:
Relevant JanusGraph query:

https://github.com/thoth-station/storages/blob/96b27cd9d536bac2b236c5aaa2655494e964593f/thoth/storages/graph/janusgraph.py#L680


fridex:
@pacospace There is a type "unparsable" in original API. We will need to check where it is used after new release.

fridex:
> @pacospace There is a type "unparsable" in original API. We will need to check where it is used after new release.

Ehm, I think only in metrics-exporter:

https://github.com/thoth-station/metrics-exporter/blob/273bb7536517be07ea35f431f74a67c4ba566237/thoth/metrics_exporter/jobs.py#L214

PR #613, opened 2019-04-18
Implemented get_all_versions_python_package method for Dgraph

pacospace:


thoth/storages/graph/dgraph.py
```diff
@@ -346,7 +346,37 @@ def get_all_versions_python_package(
...
+            {
+                f(func: has(%s)) @filter(eq(package_name, %s)%s) {
+                    package_version
+                    index_url
+                }
+            } 
+            """ % (PythonPackageVersion.get_label(), package_name, q)
+
+        result = self._query_raw(query)
+        return [[python_package['package_version'],python_package['index_url']] for python_package in result["f"][0]]
```
fridex: The signature expects tuple so explicit cast needed, also there is no need to explicitly obtain package_version and index_url as only those two are part of result.

```python
        return [tuple(python_package) for python_package in result["f"][0]]
```

thoth/storages/graph/dgraph.py
```diff
@@ -346,7 +346,37 @@ def get_all_versions_python_package(
         without_error: bool = False,
     ) -> List[Tuple[str, str]]:
         """Get all versions available for a Python package."""
-        return []
+        package_name = self.normalize_python_package_name(package_name)
```
fridex: :+1:

thoth/storages/graph/dgraph.py
pacospace: yes but python_package is still a dict ``` return [tuple(python_package.values() for python_package in result["f"]] ````

fridex: Agh, I meant:

```python
        return [tuple(python_package['package_version'], python_package['index_url']) for python_package in result["f"][0]]
```

PR #624, opened 2019-04-19
Implemented get_all_python_package_version_hashes_sha256 method for D…

pacospace:
…graph

### DIFF
--- a/thoth/storages/graph/dgraph.py
+++ b/thoth/storages/graph/dgraph.py
@@ -650,7 +650,21 @@ class GraphDatabase(StorageBase):
 
     def get_all_python_package_version_hashes_sha256(self, package_name: str, package_version: str) -> list:
         """Get hashes for a Python package per index."""
-        return []
+        package_name = self.normalize_python_package_name(package_name)
+
+        # The query requires @cascade to filter out the nodes which has index url but not artifact for that package/version
+        query = """{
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/dgraph.py
```diff
@@ -650,7 +650,20 @@ def get_python_package_version_hashes_sha256(
 
     def get_all_python_package_version_hashes_sha256(self, package_name: str, package_version: str) -> list:
         """Get hashes for a Python package per index."""
-        return []
+        package_name = self.normalize_python_package_name(package_name)
+
+        query = """{
+            f(func: has(%s)) @filter(eq(package_name, "%s") AND eq(package_version, "%s")) @cascade{
```
fridex: Why there is needed `cascade`?

pacospace: Because without @cascade that query for a certain package and version return all the index_urls accepted in Thoth including the one which has no artifact edge related to that package/version. Some packages might have more index_urls and artifacts, like tensorflow, but other could come only from one source and therefore we have to filter out with @cascade the nodes which has index url but not has_artifact edge.

fridex: Nice catch! Could you please state that in a comment so we track this behavior in sources?

pacospace: Updated

PR #630, opened 2019-04-23
Implement query for retrieving transitive dependencies

fridex:


fridex:
Just to make sure - the original query also propagated a flag about error being present (solver error). With new schema, it is possible, but we should avoid its propagation and instead have it implemented as a separate step in the adviser pipeline.

PR #631, opened 2019-04-23
Fix python_sync_analysis

pacospace:
If ```name``` key is not present in ```python_package_info["result"]```, a warning should be printed and then continue with a next item

### DIFF
--- a/thoth/storages/graph/dgraph.py
+++ b/thoth/storages/graph/dgraph.py
@@ -1346,6 +1346,11 @@ class GraphDatabase(StorageBase):
                 _LOGGER.info("Skipping error entry - %r", python_package_info)
                 continue
 
+            if not python_package_info["result"].get("name"):
+                analysis_document_id = AnalysisResultsStore.get_document_id(document)
+                _LOGGER.warning("No package name found in entry %r when syncing document %r", python_package_info, analysis_document_id)
+                continue
+
             python_package_version = PythonPackageVersion.from_properties(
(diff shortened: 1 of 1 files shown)


pacospace:
I need to rebase

fridex:
please rebase

fridex:
Please check again the PR, it removed already implemented queries.

pacospace:
updated

thoth/storages/graph/dgraph.py
```diff
@@ -1299,21 +1299,23 @@ def _python_sync_analysis_result(
...
+                    os_name=environment.os_name,
+                    os_version=environment.os_version,
+                    python_version=environment.python_version,
+                    solver_error=False,
+                    solver_error_unparseable=False,
+                    solver_error_unsolvable=False,
+                )
+                self._create_python_package_record(python_package_version, verify_index=False)
+            except Exception:  # pylint: disable=broad-exception
+               _LOGGER.exception(f"Failed to sync Python package, error is not fatal: {python_package_info!r}")
```
fridex: This will cause exceptions being reported when we detect a setup.py in a container image which does not correspond to any package (I saw these happening).

Instead, it would be better to add an if guard before creating `PythonPackageVersion`:

```python
if not python_package_info["result"].get("name"):
    _LOGGER.warning("No package name found in entry %r when syncing document %r", python_package_info, analysis_document_id)
    continue
```

and then follow with python package model creation.

PR #632, opened 2019-04-23
Implement get_python_package_tuples for Dgraph

fridex:


thoth/storages/graph/dgraph.py
```diff
@@ -559,14 +562,45 @@ def unsolved_runtime_environments(self, package_name: str, package_version: str)
         """Get unsolved runtime environment which are not connected and attached to python package version."""
         return []
 
+    async def get_python_package_tuple(self, python_package_node_id: int) -> Dict[int, tuple]:
+        """Get Python's package name, package version, package index tuple for the given package id."""
+        query = """
```
pacospace: It should be:
```
       query = '''
       {
            q(func: uid(%s)) @cascade {
                package_name
                package_version
                index_url
            }
        }
        ''' % python_package_node_id
```

fridex: aaagh :) Nice catch! :wink:

fridex: Fixed. Please review again.

PR #636, opened 2019-04-23
Implemented retrieve_unsolved_pypi_packages method for Dgraph

pacospace:


thoth/storages/graph/dgraph.py
```diff
@@ -399,13 +399,26 @@ def get_all_versions_python_package(
 
         return [tuple(python_package.values()) for python_package in result["f"]]
 
-    def retrieve_unsolved_pypi_packages(self, solver_name: str = None) -> dict:
-        """Retrieve a dictionary mapping package names to versions that dependencies were not yet resolved.
+    def retrieve_unsolved_pypi_packages(self) -> dict:
```
fridex: This signature change will break logic in graph-refresh-job:

https://github.com/thoth-station/graph-refresh-job/blob/e2eb02e2f1b934f018bb3ee2a45989fa330b743a/app.py#L105

You can do it the other way around - require solver name to be always present. I'm not sure about usage on metrics side though (it could be handy to have this method without any argument, so we count all unsolved packages).

pacospace: Updated and should I create that function as well?

fridex: I merged this functionality as that's something we use. It's up to you, I would find it interesting to see it in metrics, but that's nothing we need to do right now.

PR #639, opened 2019-04-24
Good origin dgraph

pacospace:


fridex:
Thanks @pacospace! I've based #641 based on this PR, rebased and fixed coala/pytest issues. Closing now. However, let's keep the branch available in case of any issues.

PR #661, opened 2019-05-03
Implement query for retrieving information about build-time errors

fridex:


fridex:
I will merge this and verify changes work in test environment.

PR #662, opened 2019-05-03
Created missing functions for Dgraph

pacospace:


thoth/storages/graph/dgraph.py
```diff
@@ -2032,3 +2034,53 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+                """
+            {
+                f(func: has(%s)) {
+                    c:count(uid)
+                }
+            }
+            """
+                % node_label
+            )
+            result = self._query_raw(query)
```
fridex: This can be rewritten to have one request with multiple queries to dgraph or perform multiple requests in parallel (see `_query_raw_parallel`). Let's merge this and see if we will have some perf drawbacks by doing it this way.

PR #669, opened 2019-05-05
Implement back-off for random time in case of concurent upsert writes

fridex:


fridex:
Tested in test environment.

PR #670, opened 2019-05-05
Reorganize Python package creation

fridex:
This way we minimalize upsert collisions.


fridex:
Now it works flawlessly :smiling_imp:.

PR #675, opened 2019-05-08
Introduce mechanism to avoid gRPC issues when serializing large stacks

fridex:
This fixes issue when serializing really large stacks in transitive query. An
example can be tensorflow==1.3.0rc0 which depends on other tensorflow releases
indirectly via tensoflow-tensorboard.


fridex:
I will merge this and verify changes work in test environment.

PR #699, opened 2019-05-14
Fix number of overall results

fridex:
There are present three values not just two - justification, stack
(requirements and requirements locked) and overall stack score.


goern:
regate

PR #748, opened 2019-06-14
New tests for inspection schema check before sync

pacospace:
Using https://github.com/alecthomas/voluptuous as suggested by @fridex, validation tests for the schema result of an inspection have been created. In this way we are able to check that the correct output is produced during an inspection before it can be stored on Ceph.

### DIFF
--- a/thoth/storages/inspection_schema.py
+++ b/thoth/storages/inspection_schema.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+# thoth-storages
+# Copyright(C) 2019 Fridolin Pokorny, Francesco Murdaca
+#
+# This program is free software: you can redistribute it and / or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY without even the implied warranty of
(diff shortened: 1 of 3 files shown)


thoth/storages/inspection_schema.py
```diff
@@ -0,0 +1,84 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Schema definition for inspection results."""
+
+from voluptuous import Required
+from voluptuous import Schema
+
+class Datetime(object):
```
fridex: Can you use the one defined in `result_schema.py`? Copying code is bad practice as we will need to maintain the same thing on multiple places.

pacospace: Corrected thanks!

thoth/storages/inspection_schema.py
```diff
@@ -0,0 +1,84 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Schema definition for inspection results."""
+
+from voluptuous import Required
+from voluptuous import Schema
+from result_schema import Datetime
```
fridex: I think it should be local import, otherwise this will fail in a package:

```python
from .result_schema import Datetime

```

PR #766, opened 2019-06-25
PackageAnalysisResultsStore is added

bissenbay:
This is the directory where package analyzer results will be stored on ceph.

### DIFF
--- a/thoth/storages/package_analyses.py
+++ b/thoth/storages/package_analyses.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+# thoth-storages
+# Copyright(C) 2018, 2019 Fridolin Pokorny
+#
+# This program is free software: you can redistribute it and / or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Adapter for storing package analysis results onto a persistence remote store."""
+
+from .result_base import ResultStorageBase
+
+
+class PackageAnalysisResultsStore(ResultStorageBase):
+    """Store results of package analyzes."""
+
+    RESULT_TYPE = "package-analysis"
+    SCHEMA = None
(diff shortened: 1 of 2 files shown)


thoth/storages/package_analyses.py
```diff
@@ -0,0 +1,31 @@
...
+
+"""Adapter for storing package analysis results onto a persistence remote store."""
+
+from .result_base import ResultStorageBase
+
+
+class PackageAnalysisResultsStore(ResultStorageBase):
+    """Store results of package analyzes."""
+
+    RESULT_TYPE = "package_analysis"
```
fridex: Just to be consistent here, we do dashes instead of underscores.

thoth/storages/package_analyses.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def store_document(self, package_digest: str, analysis_info: str) -> dict:
+        """Store the given document representing analysis info onto Ceph."""
+        document_id = self.ceph.store_document(analysis_info, document_id=package_digest)
+        return document_id
```
fridex: This one is already implemented in https://github.com/thoth-station/storages/blob/7b416f590b85275f1c790ccce9825856d96926f7/thoth/storages/result_base.py#L90-L96 . If you had errors due to schema validation, set `SCHEMA` class variable to `None` on this adapter.

fridex: But you should not have them if you use `print_command_result` function as described in my comment - https://github.com/thoth-station/storages/pull/766#discussion_r297234709

thoth/storages/package_analyses.py
bissenbay: Done!

thoth/storages/package_analyses.py
bissenbay: Sure. My last commit was a demonstration of how I'm storing the document. I took a look at user_api [1] how package extract stores document on ceph and it seems like we first need to schedule a package_analyzer job to Openshift and it will return the result of `print_command_result`, then I can store it to ceph. And `fetch.py` does not seem to be a place to start a connection to ceph and storing process. 
[1] https://github.com/thoth-station/user-api/blob/master/thoth/user_api/api_v1.py#L65

fridex: Exactly :+1:, results will be submitted to the result API.

thoth/storages/package_analyses.py
```diff
@@ -0,0 +1,26 @@
...
+
+"""Adapter for storing package analysis results onto a persistence remote store."""
+
+from .result_base import ResultStorageBase
+
+
+class PackageAnalysisResultsStore(ResultStorageBase):
+    """Store results of package analyzes."""
+
+    RESULT_TYPE = "package-analysis"
```
fridex: Please provide `SCHEMA`. It should match as you use `print_command_result`.

bissenbay: Done!

PR #778, opened 2019-07-01
updated adapters for storing buildlog analysis results and cache

harshad16:
updated adapters for storing build log analysis results and cache

Depends-On: #777

### DIFF
--- a/thoth/storages/__init__.py
+++ b/thoth/storages/__init__.py
@@ -23,6 +23,8 @@ from .analyses import AnalysisResultsStore
 from .analyses_by_digest import AnalysisByDigest
 from .analyses_cache import AnalysesCacheStore
 from .buildlogs import BuildLogsStore
+from .buildlogs_analyses import BuildLogsAnalysisResultsStore
+from .buildlogs_analyses_cache import BuildLogsAnalysesCacheStore
 from .ceph import CephStore
 from .dependency_monkey_reports import DependencyMonkeyReportsStore
 from .graph import GraphDatabase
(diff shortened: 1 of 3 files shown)


thoth/storages/buildlogs_analyses_cache.py
```diff
@@ -0,0 +1,30 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Adapter for retrieving and storing hashes for fast buildlog lookups."""
+
+from .ceph_cache import CephCache
+
+
+class BuildLogsAnalysesCacheStore(CephCache):
```
fridex: When this cache will be used? What is the use case here?

harshad16: I have added it, so that when build-analyzer is scheduled we store the result in cache as well and check in with the cache before scheduling.  Still a work in progress.

fridex: Nice! Good idea. :+1:

thoth/storages/__init__.py
```diff
@@ -41,4 +43,4 @@
 
 
 __name__ = "thoth-storages"
-__version__ = "0.14.4"
+__version__ = "0.14.5"
```
fridex: Versions are automatically managed by Kebechet, please do not change them manually.

harshad16: okay, cool I didn't know that. :)

PR #779, opened 2019-07-01
sync_package_analysis_documents

bissenbay:
Created model for package-analysis documents in dgraph

### DIFF
--- a/thoth/storages/graph/models.py
+++ b/thoth/storages/graph/models.py
@@ -53,6 +53,20 @@ class PackageExtractRun(VertexBase):
 
 
 @attr.s(slots=True)
+class PythonFileDigests(VertexBase):
+    """A class representing a single file digests."""
+
+    sha256 = model_property(type=str, index="exact")
+
+
+@attr.s(slots=True)
+class FileFound(ReverseEdgeBase):
+    """An edge representing a filepath to the python file found in an image."""
+
+    filepath = model_property(type=str, index="exact")
+
+
+@attr.s(slots=True)
 class Identified(ReverseEdgeBase):
     """An edge representing a package identified by a package-extract run."""
 
@@ -562,6 +576,8 @@ ALL_MODELS = frozenset(
         InspectionSoftwareStack,
         ObservedPerformance,
         PackageExtractRun,
+        PythonFileDigests,
+        FileFound,
         PiMatmul,
         ProvenanceCheckerRun,
         ProvenanceCheckerStackInput,
--- a/thoth/storages/graph/dgraph.py
+++ b/thoth/storages/graph/dgraph.py
@@ -82,6 +82,8 @@ from .models import InspectionStackInput
 from .models import InstalledFrom
 from .models import ObservedPerformance
 from .models import PackageExtractRun
+from .models import PythonFileDigests
+from .models import FileFound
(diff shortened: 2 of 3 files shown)


thoth/storages/graph/models.py
```diff
@@ -52,6 +52,21 @@ class PackageExtractRun(VertexBase):
...
 
+@attr.s(slots=True)
+class PackageAnalyzerRun(VertexBase):
+    """A class representing a single package-analyzer (package analysis) run."""
+
+    package_analysis_document_id = model_property(type=str, index="exact")
+    package_analysis_datetime = model_property(type=datetime, index="hour")
+    package_analysis_version = model_property(type=str, index="exact")
+    package_analysis_name = model_property(type=str, index="exact")
+    environment_type = model_property(type=str, index="exact")
```
bissenbay: @fridex Do we need environment_type? What else we would need to store for package analysis documents?

fridex: We will need to first finish the initial design of the package-extract output. Then, take a look at [schema](https://raw.githubusercontent.com/thoth-station/storages/master/docs/schema.png) and try to come up with some changes. We can discuss it in a meeting.

thoth/storages/graph/models.py
```diff
@@ -52,6 +52,21 @@ class PackageExtractRun(VertexBase):
     duration = model_property(type=int)
 
 
+@attr.s(slots=True)
+class PackageAnalyzerRun(VertexBase):
```
bissenbay: @fridex 1. This is how a new node is created in DGraph, right? Just curious why it's parent class is called VertexBase? 2. Should I just change the name to ```PythonFileDigests``` instead of ```PackageExtractRun``` to sync the results of Package Analyzer?

fridex: `PythonFileDigests` will point to `PythonFileDigests` so keep this one and create `PythonFileDigests`. The base class implements the actual syncing logic, these are leaf classes which define the model properties.

bissenbay:
- Created a new model for PythonFileDigests node
- Created a new model for FileFound node
- Python files found in an image are connected to PythonFileDigests node
- Updated schema.rdf
- Deleted model for PackageAnalyzerRun node and corresponding methods to it

thoth/storages/graph/models.py
```diff
@@ -52,6 +52,20 @@ class PackageExtractRun(VertexBase):
...
+    """A class representing a single file digests."""
+
+    sha256 = model_property(type=str, index="exact")
+
+
+@attr.s(slots=True)
+class FileFound(ReverseEdgeBase):
+    """An edge representing a filepath to the python file found in an image."""
+
+    filepath = model_property(type=str, index="exact", default="*")
```
fridex: No default as its always set.

bissenbay: Done!

thoth/storages/graph/models.py
bissenbay: Done

thoth/storages/graph/models.py
bissenbay: Got it!

fridex: :+1:

PR #785, opened 2019-07-04
Require non-null `index_url` and `package_name`

CermakM:
Fixes https://github.com/thoth-station/adviser/issues/303

Signed-off-by: Marek Cermak <macermak@redhat.com>

modified:   thoth/storages/graph/dgraph.py

### DIFF
--- a/thoth/storages/graph/dgraph.py
+++ b/thoth/storages/graph/dgraph.py
@@ -705,7 +705,7 @@ class GraphDatabase(StorageBase):
             PythonPackageVersion.get_label(),
             package_name,
             q,
-            "@cascade" if not only_known_index else ""
+            "@cascade" if only_known_index else ""
         )
 
         result = self._query_raw(query)
--- a/thoth/storages/graph/dgraph.py
+++ b/thoth/storages/graph/dgraph.py
@@ -667,6 +667,7 @@ class GraphDatabase(StorageBase):
         package_name: str,
         index_url: str = None,
         *,
+        only_known_index: bool = False,
         os_name: str = None,
         os_version: str = None,
         python_version: str = None,
@@ -693,12 +694,9 @@ class GraphDatabase(StorageBase):
         if index_url:
             q = q + ' AND eq(index_url, "%s")' % index_url
 
-        # required fields
-        q += " AND has(index_url) AND has(package_name)"
-
         query = """
             {
-                f(func: has(%s)) @filter(eq(package_name, %s)%s) {
+                f(func: has(%s)) @filter(eq(package_name, %s)%s) %s {
                     package_version
                     index_url
                 }
@@ -707,6 +705,7 @@ class GraphDatabase(StorageBase):
             PythonPackageVersion.get_label(),
             package_name,
             q,
+            "@cascade" if not only_known_index else ""
         )
 
         result = self._query_raw(query)
--- a/thoth/storages/graph/dgraph.py
+++ b/thoth/storages/graph/dgraph.py
@@ -693,6 +693,9 @@ class GraphDatabase(StorageBase):
         if index_url:
             q = q + ' AND eq(index_url, "%s")' % index_url
 
+        # required fields
+        q += " AND has(index_url) AND has(package_name)"
+
         query = """
             {
                 f(func: has(%s)) @filter(eq(package_name, %s)%s) {


thoth/storages/graph/dgraph.py
```diff
@@ -693,6 +693,9 @@ def get_all_versions_python_package(
         if index_url:
             q = q + ' AND eq(index_url, "%s")' % index_url
 
+        # required fields
+        q += " AND has(index_url) AND has(package_name)"
```
fridex: Is `package_name` required here? Can we parametrize injecting `has(index_url)` so we can query those which have no index assigned?

pacospace: The solution is to use @cascade so it will discard nodes which does not have both the attributes you are requesting. But I would suggest in that case, to have a parameter to add @cascade, because saying we want all versions of python packages we admit packages with no index. Therefore we can have ```only_thoth_known_index: str = False``` parameter in the query, so we don't modify previous logic, and we add @cascade which is inserted only if that parameter is set to True in the query. What do you think @fridex ?

fridex:
> The solution is to use @cascade so it will discard nodes which does not have both the attributes you are requesting. But I would suggest in that case, to have a parameter to add @cascade, because saying we want all versions of python packages we admit packages with no index. Therefore we can have only_thoth_known_index: str = False parameter in the query, so we don't modify previous logic, and we add @cascade which is inserted only if that parameter is set to True in the query. What do you think @fridex ?

Maybe the parameter is too long `only_known_indexes`? Otherwise +1.

@cascade, sorry for highlight :)

thoth/storages/graph/dgraph.py
pacospace: This should also solve https://github.com/thoth-station/adviser/issues/303 because there won't be that case in adviser. Sorry for the long name of the parameter :(

CermakM:
@pacospace @fridex if there is such parameter as `@cascade`, doesn't it make more sense to introduce something like `allow_missing_fields: bool = False` parameter which can then be used in multiple places?

fridex:
> @pacospace @fridex if there is such parameter as `@cascade`, doesn't it make more sense to introduce something like `allow_missing_fields: bool = False` parameter which can then be used in multiple places?

The fact there is `@cascade` is implementation detail - if a consumer of `GraphDatabase` does not explicitly state the property has to be present, it is not used in the query to filter out results.


> Additional question about the default value -- should it be False or True? In case of the True, we would need to modify the solver call to fix the thoth-station/adviser#303

If you meant default value of `allow_missing_fields` my comment above might answer this.

CermakM:
Okay, I might have gotten lost in the conversation, so ...
- we want both to parametrize the `has(index)` with the `only_known_index` option and
- we want to to add an option to include the `@cascade` which would be parametrized by `allow_missing_fields`.

In both cases, we don't want to modify the existing logic and therefore the `only_known_index: bool = False` and `allow_missing_fields: bool = True`.

And therefore we need to fix the solver issue by setting the `only_known_index` to `True`.

Did I get that right?

fridex:
> Okay, I might have gotten lost in the conversation, so ...
> 
> * we want both to parametrize the `has(index)` with the `only_known_index` option and
> * we want to to add an option to include the `@cascade` which would be parametrized by `allow_missing_fields`.
> 
> In both cases, we don't want to modify the existing logic and therefore the `only_known_index: bool = False` and `allow_missing_fields: bool = True`.
> 
> And therefore we need to fix the solver issue by setting the `only_known_index` to `True`.
> 
> Did I get that right?

There will be introduced a new boolean parameter `only_known_index` set to `False` by default which will not cause any logic change to the existing code base.

If `only_known_index` is set to `True` explictly, `@casade` directive is added to the query so that it filters out nodes which do not have `index_url` set.

No new `allow_missing_fields` parameter is added.

Adviser code base will adopt `only_known_index` in calls to the adjusted method which will fix https://github.com/thoth-station/adviser/issues/303 - based on the internal `@cascade` logic (nodes without `index_url` will be discarded from query results).

CermakM:
> Is package_name required here? Can we parametrize injecting has(index_url) so we can query those which have no index assigned?

So in that case, we're ignoring this comment.

fridex:
> > Is package_name required here? Can we parametrize injecting has(index_url) so we can query those which have no index assigned?
> 
> So in that case, we're ignoring this comment.
> 
> And by adding `@cascade` we implicitly require also `package_name` to be non-null.

Yes, as discussed on the internal chat thread...

thoth/storages/graph/dgraph.py
```diff
@@ -704,6 +705,7 @@ def get_all_versions_python_package(
             PythonPackageVersion.get_label(),
             package_name,
             q,
+            "@cascade" if not only_known_index else ""
```
pacospace: the default case should be the one without @cascade because the function might be implemented already in other places ```"@cascade" if only_known_index else ""```

CermakM: Ah good eye, thanks! I had there the `allow_missing_values` before and refactored incorrectly.

pacospace: No problem :) well done @CermakM :+1:

PR #807, opened 2019-07-16
Add PythonFileDigest to schema documentation

fridex:


fridex:
@bissenbay looking at the changes, it might be better to name edge "FoundFile" instead of "FileFound" (verb).

@pacospace WDYT?

bissenbay:
There is a typo on docs/schema.png. ```file_cound``` needs to be changed to ```file_found```

fridex:
> There is a typo on docs/schema.png. `file_cound` needs to be changed to `file_found`

Nice catch! I adjusted it to "found_file". Let's wait for @pacospace opinion on edge naming.

pacospace:
I agree with @fridex for the edge. I would use singular for the name of the node because there is a single hash. Otherwise LGTM! :+1:

Issue #828, opened 2019-07-22
Optimize query to Dgraph for a transitive stack


fridex:
```

2019-07-22 20:51:47,004 [1] CRITICAL root:111: Traceback (most recent call last):
--
  | File "thoth-adviser", line 711, in <module>
  | cli()
  | File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 716, in __call__
  | return self.main(*args, **kwargs)
  | File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 696, in main
  | rv = self.invoke(ctx)
  | File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 1060, in invoke
  | return _process_result(sub_ctx.command.invoke(sub_ctx))
  | File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 889, in invoke
  | return ctx.invoke(self.callback, **ctx.params)
  | File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 534, in invoke
  | return callback(*args, **kwargs)
  | File "/opt/app-root/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
  | return f(get_current_context(), *args, **kwargs)
  | File "thoth-adviser", line 397, in advise
  | limit_latest_versions=limit_latest_versions,
  | File "/opt/app-root/src/thoth/adviser/python/advise.py", line 121, in compute_on_project
  | library_usage=library_usage,
  | File "/opt/app-root/src/thoth/adviser/python/advise.py", line 82, in compute
  | return self.compute_using_pipeline(pipeline), pipeline.get_configuration()
  | File "/opt/app-root/src/thoth/adviser/python/advise.py", line 56, in compute_using_pipeline
  | for product in pipeline.conduct(count=self.count, limit=self.limit):
  | File "/opt/app-root/src/thoth/adviser/python/pipeline/pipeline.py", line 246, in conduct
  | step_context = self._initialize_stepping()
  | File "/opt/app-root/src/thoth/adviser/python/pipeline/pipeline.py", line 176, in _initialize_stepping
  | step_context = self._resolve_transitive_dependencies(direct_dependencies)
  | File "/opt/app-root/src/thoth/adviser/python/pipeline/pipeline.py", line 158, in _resolve_transitive_dependencies
  | python_version=self.project.runtime_environment.python_version,
[...]

PR #830, opened 2019-07-23
Decrease transitive query depth to address serialization issues

fridex:
Fixes: #828


fridex:
I tried to optimize this one using asyncio queue and decrease the size to 1 to run all the queries concurrently using queuing, but it turned out there was no perf gain (just implementation was much more complicated). There is missing asyncio interface to pydgraph.

PR #836, opened 2019-07-24
Fix accessing subquery result when transitive queries are broken down into multiple

fridex:


fridex:
Closing in respect to #838

PR #838, opened 2019-07-24
Optimize retrieval of transitive queries

fridex:
This way we reduce querying parts of sub-graphs which have common dependencies.


fridex:
recheck

fridex:
recheck

PR #841, opened 2019-07-24
Corrected voluptuous requirements for inspection schema

pacospace:
Corrected voluptuous requirements for inspection schema.

### DIFF
--- a/thoth/storages/inspection_schema.py
+++ b/thoth/storages/inspection_schema.py
@@ -25,23 +25,13 @@ from .result_schema import Datetime
 # Metadata about specifications produced by inspections.
 INSPECTION_SPECIFICATION_SCHEMA = Schema(
     {
-        Required("specification"): dict,
-    }
-)
-
-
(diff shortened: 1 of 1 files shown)


thoth/storages/inspection_schema.py
```diff
@@ -74,10 +61,10 @@
 INSPECTION_SCHEMA = Schema(
     {
         Required("specification"): INSPECTION_SPECIFICATION_SCHEMA,
-        Required("created"): INSPECTION_CREATED_SCHEMA,
-        Required("build_log"): INSPECTION_BUILD_LOG_SCHEMA,
+        Required("created"): str,
```
fridex: datetime?

pacospace: It's actually a string when we use the function https://github.com/thoth-station/amun-client/blob/444516e2c8df6d7e7a7caed14af47b316fd161a7/amun/lib.py#L98

PR #834, opened 2019-07-23
sync package analyzer results

bissenbay:
This is a draft on how to sync package analyzer results.
Requires schema update

PR #843, opened 2019-07-25
Update schema to include package analyzer

fridex:
Depends-On: https://github.com/thoth-station/storages/pull/834


fridex:
> @fridex As we discussed, **PythonPackageVersionEntity** node should be connected to **PackageAnalyzerRun** node on the reverse edge *_package_analyzer_input_, not **PythonPackageIndex**. Thanks!

Nice catch! Forgot to fix the schema. It should be fixed now, PTAL. :+

fridex:
BTW later we could introduce other types of files found in artifacts (future enhancement).

PR #852, opened 2019-07-30
New Dgraph function for PI

pacospace:

### DIFF
--- a/thoth/storages/graph/dgraph.py
+++ b/thoth/storages/graph/dgraph.py
@@ -108,8 +108,8 @@ from .models import UsedIn
 from .models import UsedInBuild
 from .models import UsedInJob
 from .models import UserSoftwareStack
-from .performance import ObservedPerformance
-from .performance import PERFORMANCE_MODEL_BY_NAME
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/dgraph.py
```diff
@@ -2552,3 +2552,23 @@ def get_number_of_each_vertex_in_graph(self) -> dict:
             count = result["f"][0]["c"]
             tot_nodes_per_label[node_label] = count
         return tot_nodes_per_label
+
+    def get_all_pi_per_framework_count(self, framework: str) -> dict:
+        """Retrieve dictionary with number of Performance Indicators per ML Framework in the graph database."""
+        pi_labels = [model.get_label() for model in ALL_PERFORMANCE_MODELS if "pi" in model.get_label()]
```
fridex: This looks like a dangerous part, especially as we add more performance indicators:

```python
"pi" in model.get_label()
```
If you want to check if it is a performance indicator, I would suggest to do something like:

```python
issubclass(model, PerformanceIndicatorBase)
```

pacospace: Updated thanks :)

PR #886, opened 2019-08-13
WIP Retrieve packages that are not analyzed by Package-Analyzer

bissenbay:
This is a function which queries DGraph for packages that are not analyzed by Package-Analyzer and returns a list of dictionaries, for example:
```json
[
      {
        "package_name": "pip",
        "package_version": "19.0",
        "index_url": "https://pypi.org/simple"
      },
      {
        "package_name": "pip",
        "package_version": "19.0",
        "index_url": "https://tensorflow.pypi.thoth-station.ninja/index/centos7/cuda9.2+jemalloc/simple"
      }
    ]
```

### DIFF
--- a/thoth/storages/graph/dgraph.py
+++ b/thoth/storages/graph/dgraph.py
@@ -763,6 +763,22 @@ class GraphDatabase(StorageBase):
 
         return self._postprocess_retrieve_packages(result)
 
+    def retrieve_not_analyzed_python_package_version_entities(self) -> List[dict]:
+        """Retrieve a list of package names, versions and index urls that were not analyzed yet by package-analyzer."""
+        query = """{
+            f(func: has(%s), first: 10) @filter(NOT has(%s)) {
+                package_name:package_name
+                package_version:package_version
+                index_url:index_url
+            }
+        }""" % (
+            PythonPackageVersionEntity.get_label(),
+            PackageAnalyzerInput.get_name(),
+        )
+        result = self._query_raw(query)
+
+        return result["f"]
+
     def retrieve_solved_python_packages(self, count: int = 10, start_offset: int = 0, solver_name: str = None) -> dict:
         """Retrieve a dictionary mapping package names to versions for dependencies that were already solved.
 


thoth/storages/graph/dgraph.py
```diff
@@ -763,6 +763,22 @@ def retrieve_unsolved_python_packages(self, solver_name: str) -> dict:
 
         return self._postprocess_retrieve_packages(result)
 
+    def retrieve_not_analyzed_python_package_version_entities(self) -> List[dict]:
```
bissenbay: Should we rename differently?

thoth/storages/graph/dgraph.py
```diff
@@ -763,6 +763,22 @@ def retrieve_unsolved_python_packages(self, solver_name: str) -> dict:
...
+        """Retrieve a list of package names, versions and index urls that were not analyzed yet by package-analyzer."""
+        query = """{
+            f(func: has(%s), first: 10) @filter(NOT has(%s)) {
```
bissenbay: I'm returning only the first 10 results. Should we change it or make it dynamic?

thoth/storages/graph/dgraph.py
```diff
@@ -763,6 +763,22 @@ def retrieve_unsolved_python_packages(self, solver_name: str) -> dict:
...
+                package_name:package_name
```
bissenbay: I think just adding `package_name` instead of `package_name:package_name` should be enough, but I looked at how other queries are implemented and followed the convention.

thoth/storages/graph/dgraph.py
```diff
@@ -763,6 +763,22 @@ def retrieve_unsolved_python_packages(self, solver_name: str) -> dict:
...
+                package_version:package_version
+                index_url:index_url
+            }
+        }""" % (
+            PythonPackageVersionEntity.get_label(),
+            PackageAnalyzerInput.get_name(),
+        )
+        result = self._query_raw(query)
+
+        return result["f"]
```
bissenbay: I saw most of the methods call `return self._postprocess_retrieve_packages(result)`, but as in our case the output a little different because we also returning `index_url` as it find differ. That's way I didn't use `_postprocess_retrieve_packages` .

thoth/storages/graph/dgraph.py
fridex: ah, good point... I would go with something shorter:

`retrieve_unanalyzed_python_package_versions`

thoth/storages/graph/dgraph.py
fridex: Yes, please see the linked example above.

thoth/storages/graph/dgraph.py
fridex: aliases are good when `@normalize` is used.

thoth/storages/graph/dgraph.py
fridex: Sounds good! :+1:

PR #901, opened 2019-08-21
WIP Add flag to PythonPackageEntity

bissenbay:
- Introduced a new attribute `is_present_on_index` to `PythonPackageVersionEntity`
Should fix https://github.com/thoth-station/package-analyzer/issues/31
To-do:
- Update schema

### DIFF
--- a/thoth/storages/graph/dgraph.py
+++ b/thoth/storages/graph/dgraph.py
@@ -1643,7 +1643,8 @@ class GraphDatabase(StorageBase):
             entity, _ = self.create_python_package_version_entity(
                 python_package_version.package_name,
                 python_package_version.package_version,
-                python_package_version.index_url
+                python_package_version.index_url,
+                not python_package_version.solver_error_unsolvable
             )
 
             InstalledFrom.from_properties(source=entity, target=python_package_version).get_or_create(self.client)
@@ -1656,6 +1657,7 @@ class GraphDatabase(StorageBase):
         package_name: str,
         package_version: str,
         index_url: str,
+        is_present_on_index: bool,
         *,
         only_if_package_seen: bool = False,
     ) -> Optional[Tuple[PythonPackageVersionEntity, bool]]:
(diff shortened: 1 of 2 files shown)


thoth/storages/graph/dgraph.py
```diff
@@ -1643,7 +1643,8 @@ def _create_python_package_record(
             entity, _ = self.create_python_package_version_entity(
                 python_package_version.package_name,
                 python_package_version.package_version,
-                python_package_version.index_url
+                python_package_version.index_url,
+                not python_package_version.solver_error_unsolvable
```
bissenbay: When we are creating `PythonPackageVersionEntity` if `solver_error_unsolvable` is True, then it means it is not present on index, so the attribute `is_present_on_index` for this entity would be `False`. Further I think we could change our query so only grab those `PythonPackageVersionEntity`s whose `is_present_on_index` is True and run `package-analyzer` jobs for them only.

thoth/storages/graph/dgraph.py
```diff
@@ -1656,6 +1657,7 @@ def create_python_package_version_entity(
         package_name: str,
         package_version: str,
         index_url: str,
+        is_present_on_index: bool,
```
bissenbay: Please feel free to suggest another name

thoth/storages/graph/dgraph.py
fridex: A package can be unsolvable due to other issues (e.g. Python interpreter version incompatibiliy). The `is_present_on_index` error is a subset of this issue. PTAL on this PR which will introduce a new flag in solver results which we can sync to the graph database [1].

[1] https://github.com/thoth-station/solver/pull/281

bissenbay:
https://github.com/thoth-station/solver/pull/281

bissenbay:
https://github.com/thoth-station/storages/pull/904

PR #940, opened 2019-09-10
WIP: Logic to sync inspection results

pacospace:

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -31,9 +31,11 @@ import attr
 from methodtools import lru_cache
 from sqlalchemy import create_engine
 from sqlalchemy import desc
+from sqlalchemy import func
 from sqlalchemy import tuple_
 from sqlalchemy.orm import Query
 from sqlalchemy.orm import sessionmaker
+from sqlalchemy.dialects.postgresql import insert
 from thoth.common.helpers import format_datetime
 from thoth.python import PackageVersion
 from thoth.python import Pipfile
@@ -91,7 +93,6 @@ from .models_performance import PiConv2D
 from .models_performance import PiMatmul
 from .models_performance import ALL_PERFORMANCE_MODELS, PERFORMANCE_MODEL_BY_NAME
 from collections import Counter
-from sqlalchemy import func
 
 from .sql_base import SQLBase
 from .models_base import Base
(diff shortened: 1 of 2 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -1194,6 +1197,119 @@ def create_advised_software_stack_pipfile(
...
+                build_memory = OpenShift.parse_memory_spec(document["specification"]["build"]["requests"]["memory"])
+                run_cpu = OpenShift.parse_cpu_spec(document["specification"]["run"]["requests"]["cpu"])
+                run_memory = OpenShift.parse_memory_spec(document["specification"]["run"]["requests"]["memory"])
+
+                # Convert bytes to GiB, we need float number given the fixed int size.
+                run_memory = run_memory / (1024 ** 3)
+                build_memory = build_memory / (1024 ** 3)
+
+                run_hardware_information, run_software_environment = self._runtime_environment_conf2models(
+                    document["specification"]["run"].get("requests", {}), software_environment_type="RUN", is_user=False
```
fridex: Reading this - we should improve our inspections are report runtime environment - run and build environments in the JSON output.

thoth/storages/graph/postgres.py
```diff
@@ -1194,6 +1197,119 @@ def create_advised_software_stack_pipfile(
...
+                            performance_indicator, inspection_run_id=inspection_run.id
+                        )
+
+        except Exception:
+            self._session.rollback()
+            raise
+        else:
+            self._session.commit()
+
         raise NotImplementedError
```
fridex: This can be removed now.

thoth/storages/graph/postgres.py
```diff
@@ -1194,6 +1197,119 @@ def create_advised_software_stack_pipfile(
...
+                    overall_score = document["job_log"]["stdout"].get("overall_score")
+                    if overall_score is None:
+                        _LOGGER.warning("No overall score detected in performance indicator %r", overall_score)
+
+                    performance_indicator = performance_model_class.create_from_report(self._session, document)
+
+                    if performance_indicator_name == "PiMatmul":
+                        pi_matmul, _ = PiMatmul.get_or_create(
+                            performance_indicator, inspection_run_id=inspection_run.id
+                        )
```
fridex: Is this if statement required? It looks to me the above call `performance_indicator_name` should handle creation for all the indicators.

thoth/storages/graph/postgres.py
```diff
@@ -1213,7 +1215,117 @@ def _create_python_software_stack(
...
+                        build_requests_cpu=build_cpu,
+                        build_requests_memory=build_memory,
+                        run_requests_cpu=run_cpu,
+                        run_requests_memory=run_memory,
+                        build_software_environment_id=build_software_environment.id,
+                        build_hardware_information_id=build_hardware_information.id,
+                        run_software_environment_id=run_software_environment.id,
+                        run_hardware_information_id=run_hardware_information.id,
+                    )
+                    self._session.upsert(inspection_run)
```
fridex: Does it provide `upsert`?

Anyway, I think this will not work as expected - it will create a new version which will eventually fail on `inspection_document_id` unique constraint.

I think there needs to be queried model first (just on `inspection_document_id`) and behave in two cases - dependency monkey already created it or we are creating it.

pacospace: Actually there is https://docs.sqlalchemy.org/en/13/dialects/postgresql.html?highlight=conflict#insert-on-conflict-upsert

fridex:
regate

fridex:
@pacospace still WIP?

pacospace:
I need to change the upsert logic

PR #949, opened 2019-09-10
Implement method for getting analysis metadata

fridex:


thoth/storages/graph/postgres.py
```diff
@@ -168,7 +168,23 @@ def parse_python_solver_name(solver_name: str) -> dict:
...
+            self._session.query(PackageExtractRun)
+            .filter(PackageExtractRun.analysis_document_id == analysis_document_id)
+            .with_entities(
+                PackageExtractRun.datetime,
+                PackageExtractRun.analysis_document_id,
+                PackageExtractRun.package_extract_name,
+                PackageExtractRun.package_extract_version
+            )
+        )
+        query_result = query.one()
```
pacospace: Don't we add an error if the query is empty? https://github.com/thoth-station/storages/blob/39206123de88916cbd4143b45b16796cad2bd4bc/thoth/storages/graph/dgraph.py#L303

fridex: `one()` raises an error if no results are found. We should map it to our error though.

pacospace: do you want to change it?

fridex: Adjusted in aaf5f0f :+1:

PR #952, opened 2019-09-10
Count number of performance indicators based on framework

fridex:


fridex:
@pacospace I probably miss what was the original purpose of this query. The current implementation returns number of performance indicators (regardless of performance indicator type) run. An example output would be:

```
{
  "tensorflow": 12,
  "pytorch": 9
}
```

PR #971, opened 2019-09-17
Log statistics of graph cache and memory cache if requested so

fridex:


fridex:
recheck

fridex:
recheck

fridex:
recheck

PR #998, opened 2019-09-18
New class methods for InspectionStore

pacospace:

### DIFF
--- a/thoth/storages/inspections.py
+++ b/thoth/storages/inspections.py
@@ -17,6 +17,10 @@
 
 """Adapter for persisting Amun inspection results."""
 
+
+import typing
+from typing import List, Dict
+
 from .result_base import ResultStorageBase
 from .inspection_schema import INSPECTION_SCHEMA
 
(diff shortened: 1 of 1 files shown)


thoth/storages/inspections.py
```diff
@@ -31,3 +35,23 @@ class InspectionResultsStore(ResultStorageBase):
     def get_document_id(cls, document: dict) -> str:
         """Get id under which the given document will be stored."""
         return document["inspection_id"]
+
+    def retrieve_document_ids(cls) -> List[str]:
+        """Retrieve a list of stored document from S3."""
+        return list(cls.get_document_listing())
```
fridex: This functionality is already provided by base class:

https://github.com/thoth-station/storages/blob/a214289a3c11629e40113bd3a91eeffa0120939b/thoth/storages/result_base.py#L82-L84

Which will call ceph adapter and it will return desired listing:

https://github.com/thoth-station/storages/blob/a214289a3c11629e40113bd3a91eeffa0120939b/thoth/storages/ceph.py#L60-L63

thoth/storages/inspections.py
```diff
@@ -31,3 +35,23 @@ class InspectionResultsStore(ResultStorageBase):
...
+
+    def filter_document_ids(cls, inspection_identifier: List[str]) -> Dict[str, List]:
```
fridex: I would suggest using plural as parameter is a list: inspection_identifier**s**

Issue #1021, opened 2019-09-23
Distribute alembic migrations in the package


fridex:
**Describe the bug**

As of now, we do not distribute alembic migrations in the package.

**To Reproduce**
Try to run alembic migrations in the cluster. Perform initialization of schema and see:

```python
>>> from thoth.storages import GraphDatabase
>>> graph = GraphDatabase()
>>> graph.connect()
>>> graph.initialize_schema()
  FAILED: No config file 'alembic.ini' found, or file has no '[alembic]' section
```

**Expected behavior**

Alembic migrations should be performed.

PR #1022, opened 2019-09-23
Distribute alembic migrations with thoth-storages

fridex:
Fixes: #1021


fridex:
Tested in test environment, all looks good. @pacospace PTAL, I will solve coala issues.

PR #1023, opened 2019-09-23
Introduce a check to verify the current database schema is up2date

fridex:
Depends-On: #1022


goern:
Can we have this as part of the readyness probe of services, so that a new deployment blocks until the database schema has the correct version? @fridex @pacospace @harshad16

fridex:
> Can we have this as part of the readyness probe of services, so that a new deployment blocks until the database schema has the correct version? @fridex @pacospace @harshad16

I think we should do this not only for services but for all the components which occur in the cluster and talk to the database - ideally, during release management, execute this as one of the verification steps in the integration testsuite.

PR #1044, opened 2019-09-26
New query: count software stacks per type

pacospace:

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -1401,6 +1401,22 @@ class GraphDatabase(SQLBase):
 
         return software_stack
 
+    def python_software_stack_count(
+        self,
+        software_stack_type: str,
+        unique: bool = False,
+    ) -> List[str]:
+        """Get number of Python software stacks available filtered by type."""
+        query = (
+            self._session.query(PythonSoftwareStack.software_stack_type)
+            .filter(PythonSoftwareStack.software_stack_type == software_stack_type)
+        )
+
+        if unique:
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -1400,6 +1400,23 @@ def _create_python_software_stack(
...
+        query = (
+            self._session.query(PythonSoftwareStack.software_stack_type)
+            .filter(PythonSoftwareStack.software_stack_type == software_stack_type)
+            .offset(start_offset)
+            .limit(count)
+        )
+        if unique:
+            return len([item[0] for item in query.distinct().all()])
+
+        return len([item[0] for item in query.all()])
```
fridex: As the database grows, this will always return `count`. For counting queries we should use `.count()` to let database count entries for us (so we reduce data sending/receive overhead)

thoth/storages/graph/postgres.py
```diff
@@ -1401,6 +1401,26 @@ def _create_python_software_stack(
...
+        start_offset: int = 0,
+        count: int = 1000,
+        unique: bool = False,
+    ) -> List[str]:
+        """Get number of Python software stacks available filtered by type."""
+        query = (
+            self._session.query(PythonSoftwareStack.software_stack_type)
+            .filter(PythonSoftwareStack.software_stack_type == software_stack_type)
+            .offset(start_offset)
+            .limit(count)
```
fridex: `offset` and `limit` are not needed

pacospace:
Updated! Thanks :beers:

PR #1049, opened 2019-09-27
[WIP] New query for number of unanalyzed python pacakges query

pacospace:

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -555,6 +555,41 @@ class GraphDatabase(SQLBase):
 
         return query.count()
 
+    def retrieve_unanalyzed_python_packages_versions_count(self) -> int:
+        """Retrieve number of unanalyzed Python packages/versions"""
+        subquery = (
+            self._session.query(PackageAnalyzerRun)
+            .join(PythonPackageVersionEntity)
+            .join(PythonPackageIndex)
+            .with_entities(
+                PythonPackageVersionEntity.package_name,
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -555,6 +555,41 @@ def retrieve_solved_python_packages_count(self, solver_name: str = None) -> int:
...
+
+        query_result = (
+            self._session.query(PythonPackageVersionEntity)
+            .join(PythonPackageIndex)
+            .filter(
+                tuple_(
+                    PythonPackageVersionEntity.package_name,
+                    PythonPackageVersionEntity.package_version,
+                    PythonPackageIndex.url,
+                ).notin_(subquery)
```
fridex: There will be required a check for `index_url` not being `None` - packages which we detect in container image analysis should be excluded from counting.

Also, the query is copy of `retrieve_unanalyzed_python_package_versions` (which has the bug mentioned above as well :angel:). Can we place code at one place as done in `_construct_unsolved_python_packages_query` and relevant `retrieve_unsolved_python_packages`, `retrieve_unsolved_python_packages_count` methods?

pacospace: Yes thanks, I created PR and it is a WIP, I need to solve consistency also to be sure we are counting exactly the correct things and I will take care of that as well :) :beers:

pacospace:
No more relevant, substituted by: https://github.com/thoth-station/storages/pull/1093

PR #1053, opened 2019-09-27
Add duration to the result schema

fridex:
... so that results pass and are correctly stored on Ceph.

Fixes: #1052


fridex:
recheckk

fridex:
recheck

fridex:
recheck

fridex:
This looks like some update broke the testsuite?!

PR #1074, opened 2019-10-02
Add logic for syncing marker and extra

fridex:


fridex:
Added missing migration - the change is backwards compatible, so dropping dependency on https://github.com/thoth-station/solver/pull/312

PR #1093, opened 2019-10-09
Created and Updated queries for analyzed packages

pacospace:
- Created functions for analyzed packages according to new naming convention.
- Added functions for solved packages with error

Depends-On: https://github.com/thoth-station/storages/pull/1104


pacospace:
thoth-coala failed because we are above 4000 lines now, https://github.com/thoth-station/storages/blob/534d1cc3870c90ad9e8ca2b686b17dc97cf2a231/.coafile#L7

thoth/storages/graph/postgres.py
```diff
@@ -908,24 +1039,111 @@ def get_unsolved_python_packages_all_versions(
...
-        for item in query:
-            if item[0] not in query_result.keys():
+        for item in result:
+            if item[0] not in query_result:
                 query_result[item[0]] = []
+            query_result[item[0]].append((item[1], item[2]))
+
+        unsolved = self._get_unsolved_python_package_versions_count_edge_cases()
+
+        if unsolved and count > 0:
```
fridex: This is kind-of weird behavior

pacospace: I adjusted it, WDYT?

pacospace: In this way, it is more clear, we are splitting in two and we have +- 1 in count.

thoth/storages/graph/postgres.py
```diff
@@ -979,9 +1206,20 @@ def get_unsolved_python_package_versions_count(
...
-        query = query.all()
+        result = query.all()
 
-        return {(item[0], item[1], item[2]): item[3] for item in query}
+        unsolved = []
+        if len(result) < count:
+            unsolved = self._get_unsolved_python_package_versions_count_edge_cases(count=count - len(result))
+
+        query_result = {}
+        for item in result + unsolved:
```
fridex: Not to create a list in memory for this (other places as well):

```
for item in itertools.chain(result. unsolved):
```

thoth/storages/graph/postgres.py
```diff
@@ -979,9 +1206,20 @@ def get_unsolved_python_package_versions_count(
...
+            if (item[0], item[1], item[2]) not in query_result.keys():
+                query_result[(item[0], item[1], item[2])] = item[3]
+            else:
+                query_result[(item[0], item[1], item[2])] += item[3]
```
fridex: append

thoth/storages/graph/postgres.py
```diff
@@ -979,9 +1206,20 @@ def get_unsolved_python_package_versions_count(
...
+        result = query.all()
 
-        return {(item[0], item[1], item[2]): item[3] for item in query}
+        unsolved = []
+        if len(result) < count:
+            unsolved = self._get_unsolved_python_package_versions_count_edge_cases(count=count - len(result))
+
+        query_result = {}
+        for item in result + unsolved:
+            if (item[0], item[1], item[2]) not in query_result.keys():
```
fridex:
```python
if item not in query_result
```

thoth/storages/graph/postgres.py
```diff
@@ -775,16 +892,108 @@ def _construct_solved_query(
...
+            case_3 = case_3.filter(PythonPackageVersionEntity.package_name == package_name)
+
+        if package_version:
+            case_3 = case_3.filter(PythonPackageVersionEntity.package_version == package_version)
+
+        if distinct:
+            case_3 = case_3.distinct()
+
+        if not is_count:
+            case_3 = []
```
fridex: Why is this needed?

thoth/storages/graph/postgres.py
```diff
@@ -1266,14 +1482,80 @@ def _get_unsolved_python_package_edge_cases(
...
+
+        unsolved = []
+        if len(result) < count:
+            unsolved = self._get_unsolved_python_package_versions_edge_cases(
+                package_name=package_name,
+                package_version=package_version,
+                count=count - len(result)
+            )
+
+        result = result + unsolved
```
fridex:
```
result.extend(unsolved)
```

thoth/storages/graph/postgres.py
```diff
@@ -1308,172 +1590,871 @@ def get_unsolved_python_package_versions_count_all(
...
+        if len(result) < count:
+            unsolved = self._get_unsolved_python_package_versions_edge_cases(
+                package_name=package_name,
+                package_version=package_version,
+                distinct=distinct,
+                is_count=True
+            )
 
-        total_count = query + len(unsolved)
+        total_count = result + len(unsolved)
```
fridex: For performance, it would be good to introduce a parameter to `_get_unsolved_python_package_versions_edge_cases` which will trigger counting (its a private method, so it should be ok).

thoth/storages/graph/postgres.py
pacospace: do you mean `update`?

thoth/storages/graph/postgres.py
pacospace: not needed indeed :) nice catch! Thanks :D

thoth/storages/graph/postgres.py
```diff
@@ -979,9 +1210,21 @@ def get_unsolved_python_package_versions_count(
...
+        result = query.all()
+
+        unsolved = []
+        if len(result) < count:
+            unsolved = self._get_unsolved_python_package_versions_count_edge_cases(count=count - len(result))
+
+        query_result = {}
 
-        return {(item[0], item[1], item[2]): item[3] for item in query}
+        for item in itertools.chain(result. unsolved):
```
fridex: A comma instead of a dot.

thoth/storages/graph/postgres.py
```diff
@@ -1042,16 +1290,21 @@ def get_unsolved_python_package_versions_count_per_index(
...
 
         query_result = {}
         query_result[index_url] = {}
 
-        for item in query:
-            if (item[0], item[1]) not in query_result[index_url].keys():
-                query_result[index_url][(item[0], item[1])] = item[3]
-            else:
-                query_result[index_url][(item[0], item[1])] += item[3]
+        for item in itertools.chain(result. unsolved):
```
fridex: A comma instead of a dot.

thoth/storages/graph/postgres.py
pacospace: No problem :) Thanks!

thoth/storages/graph/postgres.py
pacospace: Thanks!

PR #1100, opened 2019-10-10
Fix linkage of artifacts in Python package version entities

fridex:
We should link to just entity, as artifacts are not specific to OS
name/version/python version in the current design.


fridex:
@pacospace could you please commit the adjusted diagram in docs? :)

PR #1101, opened 2019-10-10
Consistenly sync index_url and package_version

bissenbay:
Depends-On: https://github.com/thoth-station/solver/pull/329

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3429,8 +3429,8 @@ class GraphDatabase(SQLBase):
             for error_info in document["result"]["errors"]:
                 # Normalized in `_create_python_package_version'.
                 package_name = error_info.get("package_name") or error_info["package"]
-                package_version = error_info["version"]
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -3429,8 +3429,8 @@ def sync_solver_result(self, document: dict) -> None:
             for error_info in document["result"]["errors"]:
                 # Normalized in `_create_python_package_version'.
                 package_name = error_info.get("package_name") or error_info["package"]
-                package_version = error_info["version"]
-                index_url = error_info["index"]
+                package_version = error_info.get("package_version")
+                index_url = error_info.get("index_url")
```
fridex: With this, we will not be consistent with older versions. Can this be switched to something like this?

```python
                package_version = error_info.get("package_version", error_info["version"])
                index_url = error_info.get("index_url", error_info["index"])
```

bissenbay: Sure, thanks!

Issue #1123, opened 2019-10-21
Multiple heads error in alembic migrations


fridex:
**Describe the bug**

As we merged two PRs (https://github.com/thoth-station/storages/pull/1110 and https://github.com/thoth-station/storages/pull/1102) that affect database schema and rely on the same previous database schema, our branch heads diverged.

**To Reproduce**

Try to run alembic migrations:

```
$ PYTHONPATH=. pipenv run alembic upgrade head
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
ERROR [alembic.util.messaging] Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
  FAILED: Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
```

**Expected behavior**

Alembic migrations should succeed.

**Additional context**

```
 PYTHONPATH=. pipenv run alembic history
184e698fe2c2 -> 74f68dabbb46 (head), Sync python interpreters
184e698fe2c2 -> e05d6d95ea95 (head), Added PythonPackageMetadata table
a8a4748e85cb -> 184e698fe2c2 (branchpoint), Artifact should link just to entity
331cbc965699 -> a8a4748e85cb, HasArtifact linked to PythonPackageVersionEntity
079e1d92f9d4 -> 331cbc965699, Relax constraints for ExternalSoftwareEnvironment and SoftwareEnvironment for PackageExtractRun
650a13fd0a08 -> 079e1d92f9d4, add is_external to PackageExtractRun
9a8ce1e52efb -> 650a13fd0a08, Robust Syncs
f60593780969 -> 9a8ce1e52efb, Added ExternalSoftwareEnvironment, ExternalHardwareInformation, constraints for syncs
1ee67d9a7861 -> f60593780969, Add markers and extra needed for resolver
9168881dd125 -> 1ee67d9a7861, Enable nullable limit_latest_versions
<base> -> 9168881dd125, Initial schema
```

fridex:
CC: @bissenbay @pacospace

PR #1124, opened 2019-10-21
Fix multiple heads present

fridex:
Fixes: #1123


fridex:
> Great! Also in line 4 needs to be updated :)

Thanks! Adjusted :)

PR #1134, opened 2019-10-21
[WIP] Get internal software & hardware environments

bissenbay:


thoth/storages/graph/postgres.py
```diff
@@ -2970,6 +2970,66 @@ def python_package_index_listing(self, enabled: bool = None) -> list:
 
         return [{"url": item[0], "warehouse_api_url": item[1], "verify_ssl": item[2]} for item in query.all()]
 
+    def internal_hardware_environments_listing(self) -> list:
+        """Get a list of hardware environments registered in the graph database."""
+        query = self._session.query(
+            HardwareInformation.cpu_cores,
```
fridex: You can directly issue:

```python
result = self._session.query(HardwareInformation).all()
```

and subsequently transform models returned using:

```python
return [model.to_dict() for model in result]
```

thoth/storages/graph/postgres.py
```diff
@@ -2970,6 +2970,66 @@ def python_package_index_listing(self, enabled: bool = None) -> list:
...
+                "gpu_cores": item[6],
+                "gpu_memory_size": item[7],
+                "gpu_model_name": item[8],
+                "gpu_vendor": item[9],
+                "ram_size": item[10]
+
+            } for item in query.all()
+        ]
+
+    def internal_software_environments_listing(self) -> list:
```
fridex: I'm not sure if this corresponds to our naming schema for queries. I would expect one method here which accepts a parameter distinguishing whether the returned sw/hw is internal or external.

@pacospace PTAL

pacospace: I was thinking 

```
    def get_software_environments_listing(self, is_external: bool = False) -> List[Dict]:
        """Get a list of software environments registered in the graph database."""
        if is_external:
            software_environment = ExternalSoftwareEnvironment
        else:
            software_environment = SoftwareEnvironment

        query = self._session.query(
            software_environment.cuda_version,
            software_environment.environment_name,
            software_environment.environment_type,
            software_environment.image_name,
            software_environment.image_sha,
            software_environment.os_name,
            software_environment.os_version,
            software_environment.python_version
        )

        return [
            {
                "cuda_version": item[0],
                "environment_name": item[1],
                "environment_type": item[2],
                "image_name": item[3],
                "image_sha": item[4],
                "os_name": item[5],
                "os_version": item[6],
                "python_version": item[7]

            } for item in query.all()
        ]
```

This could be done also for the hardware, so we would have a single function and reuse the same flag `is_external` we have in storages

bissenbay: Thanks!

thoth/storages/graph/postgres.py
bissenbay: Thanks!

bissenbay:
@fridex This is I mentioned in the meeting.

PR #1139, opened 2019-10-22
Symbol-API

KPostOffice:
# Description
Add an API for getting symbols associated with both python packages and images.

This is required for adding a new step to advise.

Both new functions should return a list of symbols.


thoth/storages/graph/postgres.py
```diff
@@ -2894,6 +2894,60 @@ def get_python_package_version_hashes_sha256(
...
+            .filter(PackageExtractRun.cuda_version == cuda_version)
+            .with_entities(
+                PackageExtractRun.versioned_symbols
+            )
+            .limit(1)
+        )
+        query_result = query.fetch()
+
+        if query_result is None:
+            raise NotFoundError(f"No records found for the given runtime environment.")
```
fridex: Please be more descriptive in the exception - state parameters that were used in the query.

thoth/storages/graph/postgres.py
```diff
@@ -2894,6 +2894,60 @@ def get_python_package_version_hashes_sha256(
         )
         return [item[0] for item in query.all()]
 
+    # BOOKMARK
```
fridex: I think this can be removed.

thoth/storages/graph/postgres.py
```diff
@@ -2894,6 +2894,60 @@ def get_python_package_version_hashes_sha256(
...
+    def get_python_package_required_symbols(
+        self, package_name: str, package_version: str, index_url: str = None
```
fridex: I would probably require `index_url` to be set. Our own builds may require other symbols than the upstream ones.

thoth/storages/graph/postgres.py
```diff
@@ -2894,6 +2894,68 @@ def get_python_package_version_hashes_sha256(
...
         return [item[0] for item in query.all()]
 
+    def get_python_package_required_symbols(
+        self, package_name: str, package_version: str, index_url: str
+    ) -> List[str]:
+        """Get required symbols for a Python package in a specified version."""
+        package_name = self.normalize_python_package_name(package_name)
+        package_version = self.normalize_python_package_version(package_version)
+
+        if index_url is not None:
```
fridex: `index_url` will never be `None` based on the function signature. You can drop the `else` branch.

thoth/storages/graph/postgres.py
```diff
@@ -2894,6 +2894,68 @@ def get_python_package_version_hashes_sha256(
...
+        else:
+            query = self._session.query(PythonPackageVersion)
+
+        query = (
+            query.filter(PythonPackageVersion.package_name == package_name)
+            .filter(PythonPackageVersion.package_version == package_version)
+            .join((HasArtifact, PythonPackageVersion.python_artifacts))
+            .join((PythonArtifact, HasArtifact.python_artifact))
+            .with_entities(PythonArtifact.versioned_symbols)
+            .limit(1)
```
fridex: Why do you use `disticnt` here (instead of `limit` as before)?

thoth/storages/graph/postgres.py
```diff
@@ -2894,6 +2894,68 @@ def get_python_package_version_hashes_sha256(
...
+        )
+
+        result = query.all()
+        if (len(result) == 0):
```
fridex: Brackets not needed in Python.

thoth/storages/graph/postgres.py
```diff
@@ -2894,6 +2894,68 @@ def get_python_package_version_hashes_sha256(
...
+            self._session.query(PackageExtractRun)
+            .filter(PackageExtractRun.analysis_document_id == environment_name)
+            .filter(PackageExtractRun.os_name == os_name)
+            .filter(PackageExtractRun.os_version == os_version)
+            .filter(PackageExtractRun.python_version == python_version)
+            .filter(PackageExtractRun.cuda_version == cuda_version)
+            .with_entities(
+                PackageExtractRun.versioned_symbols
+            )
+            .limit(1)
```
fridex: `distinct` instead of `limit`?

goern:
recheck

thoth/storages/graph/postgres.py
```diff
@@ -2957,6 +2957,68 @@ def get_python_package_version_hashes_sha256(
...
+        )
+        query_result = query.fetch()
```
pacospace: Maybe `query.all()`? or if you want first one, `query.first()`?

pacospace:
It needs to be reopened @KPostOffice, I accidentally put approved, instead of comment in the review :/

PR #1141, opened 2019-10-23
Drop cache support

fridex:
As PostgreSQL seems to be performing well, let's drop graph cache support.

There are planned changes to queries which would respect extras to be fully
Python compliant - maintainig cache code does not make sense.


fridex:
I've already archived https://github.com/thoth-station/graph-cache-refresh-job/ (which is not part of deployment now).

PR #1143, opened 2019-10-23
Add support for extras in the Python package dependencies retrieval query

fridex:


fridex:
This will break integration with the current adviser, but it is about to be replaced anyway...

PR #1147, opened 2019-10-23
Handle issues in a better way

bissenbay:
Related-To: https://github.com/thoth-station/cve-update-job/issues/218

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -4094,7 +4094,7 @@ class GraphDatabase(SQLBase):
                 cve, _ = CVE.get_or_create(
                     self._session, cve_id=record_id, version_range=version_range, advisory=advisory, cve_name=cve
                 )
-                index = self._session.query(PythonPackageIndex).filter_by(url=index_url).one()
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -4495,7 +4495,7 @@ def sync_package_analysis_result(self, document: dict) -> None:
             self._session.commit()
 
     def _get_or_create_python_package_index(
-        self, index_url: str, only_if_enabled: bool = True
+        self, index_url: str, only_if_enabled: bool = False
```
fridex: I meant replacing line:

```python
    index = self._session.query(PythonPackageIndex).filter_by(url=index_url).one()
```

with:

```python
    index = self._get_or_create_python_package_index(index_url, only_if_enabled=False)
```

in `create_python_cve_record` method.

goern:
Will it fix 218?

fridex:
> Will it fix 218?

Yes, CVEs will be synced into the database and the PyPI index will be disabled until we run init-job.

goern:
Do we need a new release due to this fix?

PR #1159, opened 2019-10-25
Fix index url key in new solvers implementation

fridex:


fridex:
> This `unsolvable["index"]` might still through the error. @fridex

Nice catch, same issue as in https://github.com/thoth-station/storages/pull/1160

fridex:
See #1161

PR #1178, opened 2019-10-27
HasArtifact is linked with PythonPackageVersionEntity table

fridex:
This fixes error during retrieval of hashes when there is queried
PythonPackageVersion table, but this table has no relation to HasArtifact
table.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -2942,16 +2942,16 @@ class GraphDatabase(SQLBase):
             query = (
                 self._session.query(PythonPackageIndex)
                 .filter(PythonPackageIndex.url == index_url)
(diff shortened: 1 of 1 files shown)


pacospace:
> This fixes error during retrieval of hashes when there is queried
> PythonPackageVersion table, but this table has no relation to HasArtifact
> table.

PPVE contains PPV, but PPV actually should be related as well, because we extract PythonArtifacts also with solvers, therefore there is a missing link in PythonArtifact table with PPV I think. https://github.com/thoth-station/storages/blob/88afe6106e542da6a172fb6f4affb8bf88816d20/thoth/storages/graph/postgres.py#L4706. We found out about a new bug, nice! :1st_place_medal:

PR #1183, opened 2019-10-29
Issue warning when syncing and database is not up2date

fridex:


fridex:
@pacospace or we could issue this warning in the `connect` method so we get this warning on each connection establishement.

fridex:
Possibly replaced by #1184

fridex:
> > @pacospace or we could issue this warning in the `connect` method so we get this warning on each connection establishment.
> 
> I would choose the check at each connection  , because it is not only for syncing in general. For example there could be a problem for a new query using a new model, but if the schema is not up2date there would be an error for table missing. Thanks :)

On the same page, sir :+1: :1st_place_medal: 

Closing with respect to #1183

PR #1185, opened 2019-10-29
State maintainer and project url in setup.py

fridex:


setup.py
```diff
@@ -96,6 +96,9 @@ def run_tests(self):
     entry_points={
         'console_scripts': ['thoth-storages=thoth.storages.cli:cli']
     },
+    url='https://github.com/thoth-station/storages',
```
pacospace: I'm not sure about that person :P

PR #1189, opened 2019-10-29
Issue warning if the database schema is not initialized yet in connect

fridex:
Fixes: https://github.com/thoth-station/management-api/issues/269


fridex:
This introduced a weird chicken-egg problem. We cannot connect and initialize schema because each time we try to connect, we fail. @harshad16 please review as hotfix and we can issue a new patch request.

harshad16:
Just observed the chicken-egg problem, thanks @fridex  for the fix :+1:

PR #1196, opened 2019-10-30
Fix retrieval of Python digests query

fridex:
* no need to have pagination
* fix relation query

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3004,14 +3004,12 @@ class GraphDatabase(SQLBase):
 
         return [cve.to_dict() for cve in result]
 
-    def get_python_package_version_hashes_sha256_all(
+    def get_python_package_hashes_sha256(
         self,
-        package_name: str = None,
-        package_version: str = None,
-        index_url: str = None,
+        package_name: str,
+        package_version: str,
+        index_url: str,
         *,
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -3006,12 +3006,10 @@ def get_python_cve_records_all(self, package_name: str, package_version: str) ->
 
     def get_python_package_version_hashes_sha256_all(
```
pacospace: If we change it in this way, the query convention is not respected, because with `all` in the name, we should get all python_package_version hashes in the database if no filter are provided, therefore a List[str]. Then we can filter for a specific package version. With these changes, package_name, package_version and index_url have to be provided. If we want to follow convention, it should be called `get_python_artifact_hashes_sha256_per_python_package_version()`, so we follow `get_<objects>_per_<attribute_name>()` but it should return in this case Dict[Tuple[str, Any, Any], List[str]]

pacospace: We could also rename it  `get_python_package_version_hashes_sha256` and accept the changes requested :)

fridex: :+1: Seems like the naming convention works! :+1: 

Adjusted to the latter one - `get_python_package_hashes_sha256`.

Issue #1174, opened 2019-10-26
[RFE] use sqlalchemy-stubs


fridex:
https://github.com/dropbox/sqlalchemy-stubs

PR #1197, opened 2019-10-30
Start using mypy in strict mode

fridex:
Fixes: #1174


fridex:
> LGTM! Thanks

We will need to fix issues first to get this in.

PR #1199, opened 2019-10-30
Correct errors for mypy

pacospace:


thoth/storages/result_base.py
```diff
@@ -51,8 +51,8 @@ def __init__(
         explicitly.
         """
         assert (
-            self.RESULT_TYPE is not None
-        ), "Make sure you define RESULT_TYPE in derived classes to distinguish between adapter type instances."
+            self.RESULT_TYPE is not ""
```
fridex: Some linters might complain here, you can do just `not self.RESULT_TYPE`.

Issue #1219, opened 2019-11-04
extend sync_document() to read from local file


goern:
**Is your feature request related to a problem? Please describe.**
no

**Describe the solution you'd like**
sync_document() should be able to read a file from the local directory.

**Additional context**
n/a

PR #1220, opened 2019-11-04
Sync documents from a local directory if requested

fridex:
Fixes: #1219

### DIFF
--- a/thoth/storages/sync.py
+++ b/thoth/storages/sync.py
@@ -18,8 +18,12 @@
 """Routines for syncing data from Ceph into graph database."""
 
 import logging
+import json
+from pathlib import Path
 from typing import Dict
 from typing import Tuple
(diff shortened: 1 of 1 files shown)


thoth/storages/sync.py
```diff
@@ -41,15 +45,25 @@
...
+        )
+
     if not graph:
         graph = GraphDatabase()
         graph.connect()
 
-    adviser_store = AdvisersResultsStore()
-    adviser_store.connect()
+    if not is_local:
+        adviser_store = AdvisersResultsStore()
```
goern: Is the scope of adviser_store only within this if?!

fridex: Nope, the variable will exist if the statement is evaluated as true (not always nice, I know...).

PR #1224, opened 2019-11-04
New models for PythonPackage Metadata that have multiple values

pacospace:
- Added multi-part keys metadata modules
- Added Migrations Schema
- Updated sync logic
- Updated Schema docs 
- Tested syncs


thoth/storages/graph/models.py
```diff
@@ -371,7 +371,7 @@ class CVE(Base, BaseExtension):
     cve_id = Column(String(256), nullable=True)
     version_range = Column(String(256), nullable=True)
 
-    python_package_versions = relationship("HasVulnerability", back_populates="cve")
+    python_package_version_entities = relationship("HasVulnerability", back_populates="cve")
```
fridex: Nice catch! :+1:

thoth/storages/graph/postgres.py
```diff
@@ -4727,17 +4726,42 @@ def sync_solver_result(self, document: dict) -> None:
...
+                            python_package_classifier, _ = PythonPackageClassifier.get_or_create(
+                                self._session,
+                                classifier=classifier,
+                            )
+                            HasClassifier.get_or_create(
+                                self._session,
+                                python_package_classifier_id=python_package_classifier.id,
+                                python_package_metadata_id=package_metadata.id,
+                            )
+                        importlib_metadata.pop("Classifier", None)
```
pacospace: WDYT about this way @fridex ?

fridex: +1, these items have to be removed from the dictionary due to warning produced later on. The code could be simplified to:

```python
                        for classifier in importlib_metadata.pop("Classifier", []):
                            python_package_classifier, _ = PythonPackageClassifier.get_or_create(
                                self._session,
                                classifier=classifier,
                            )
                            HasClassifier.get_or_create(
                                self._session,
                                python_package_classifier_id=python_package_classifier.id,
                                python_package_metadata_id=package_metadata.id,
                            )
```

pacospace:
recheck

thoth/storages/graph/postgres.py
```diff
@@ -4677,6 +4695,106 @@ def _get_or_create_python_package_index(
...
+                distutils_type = MetadataDistutilsTypeEnum.REQUIRED.value
+
+            elif "Provides-Dist":
+                distutils_type = MetadataDistutilsTypeEnum.PROVIDED.value
+
+            elif "Obsoletes-Dist":
+                distutils_type = MetadataDistutilsTypeEnum.OBSOLETE.value
+
+            else:
+                _LOGGER.warning("No distutils key registered for %r", dist_key)
```
fridex: This will cause the `distutils_type` not being defined on the model created bellow causing a crash during execution. You can directly raise an exception here.

pacospace: I'm gonna create a specific exception then, thanks :)

PR #1254, opened 2019-11-11
added more builds that need to be triggered

goern:
Signed-off-by: Christoph Görn <goern@redhat.com>

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,13 +1,14 @@
 - project:
     check:
-      queue: thoth-station/core
+      queue: thoth-station/storages
       jobs:
         - thoth-coala
         - thoth-pytest
     gate:
-      queue: thoth-station/core
+      queue: thoth-station/storages
       jobs:
         - thoth-coala
+        - thoth-pytest
     release:
       jobs:
         - upload-pypi-sesheta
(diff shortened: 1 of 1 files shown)


harshad16:
last week I was thinking of adding this too, however didn't move forward, as once the new release is made for storages, the kebechet instance would try to update the other components and when the update pr is merged in other repo, they would be triggered as well. 
Definitely this is a faster way and might provide reassurance.

harshad16:
@goern I m confused about the `queue` instance though, is calling storages while testing storages a good practice?

goern:
a) why is zuul not gating?!
b) I think this will build all changes between releases

goern:
> @goern I m confused about the `queue` instance though, is calling storages while testing storages a good practice?

chk https://zuul-ci.org/docs/zuul/user/config.html#attr-project.%3Cpipeline%3E.queue

maybe we should put it back in core queue...

.zuul.yaml
```diff
@@ -1,13 +1,14 @@
 - project:
     check:
-      queue: thoth-station/core
+      queue: thoth-station/storages
       jobs:
         - thoth-coala
         - thoth-pytest
     gate:
-      queue: thoth-station/core
+      queue: thoth-station/storages
```
harshad16: :)

goern:
regate

harshad16:
> regate
this would, work when we approve it again.

harshad16:
> a) why is zuul not gating?!

I guess, we figured that part out. Now it would work.

> b) I think this will build all changes between releases

Image build trigger for all of those repos for every PR merge of storages doesn't add upto me, as they would only get to know about these changes when there is a release of storages. 

@pacospace  and @fridex  PTAL

harshad16:
@fridex  @pacospace  PTAL :)

harshad16:
recheck

PR #1277, opened 2019-11-15
Dispose engine on connect issues

fridex:


thoth/storages/graph/postgres.py
```diff
@@ -190,6 +190,19 @@ def connect(self):
         # We do not use connection pool, but directly talk to the database.
         self._engine = create_engine(self.construct_connection_string(), echo=echo, pool_pre_ping=True)
         self._session = sessionmaker(bind=self._engine)()
+        try:
+            self._engine = create_engine(self.construct_connection_string(), echo=echo, poolclass=NullPool)
+            self._session = sessionmaker(bind=self._engine)()
+        except Exception:
+            # Drop engine and session in case of anny connection issues so is_connected behaves correctly.
```
pacospace: one typo in `any` otherwise LGTM

fridex: Nice catch! Fixed :+1: I wonder whether we should do the same on disconnect (probably yes).

PR #1297, opened 2019-11-19
Introduce sorting type in queries

pacospace:

### DIFF
--- a/thoth/storages/graph/enums.py
+++ b/thoth/storages/graph/enums.py
@@ -63,3 +63,10 @@ class MetadataDistutilsTypeEnum(Enum):
     # rarely used fields
     PROVIDED = "PROVIDED"
     OBSOLETE = "OBSOLETE"
+
+
+class QuerySortTypeEnum(Enum):
+    """Class for the requirements format."""
+
+    PACKAGE_NAME = "package_name"
+    PACKAGE_VERSION = "package_version"
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -127,10 +127,12 @@ from collections import Counter
 
 from .sql_base import SQLBase
 from .models_base import Base
+from .query_result_base import QueryResult
 from .enums import EnvironmentTypeEnum
 from .enums import SoftwareStackTypeEnum
 from .enums import InspectionSyncStateEnum
 from .enums import MetadataDistutilsTypeEnum
+from .enums import QuerySortTypeEnum
 
 from ..analyses import AnalysisResultsStore
 from ..dependency_monkey_reports import DependencyMonkeyReportsStore
@@ -149,6 +151,7 @@ from ..exceptions import AlreadyConnected
 from ..exceptions import DatabaseNotInitialized
 from ..exceptions import SolverNameParseError
 from ..exceptions import DistutilsKeyNotKnown
+from ..exceptions import SortTypeQueryError
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -3287,7 +3290,8 @@ class GraphDatabase(SQLBase):
         os_version: str = None,
         python_version: str = None,
         distinct: bool = False,
-    ) -> Dict[str, int]:
+        sort_by: QuerySortTypeEnum = None
+    ) -> QueryResult:
         """Retrieve number of versions per Python package name in Thoth Database.
 
         Examples:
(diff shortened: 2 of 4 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -3257,7 +3258,8 @@ def get_python_package_versions_all_count(
         os_version: str = None,
         python_version: str = None,
         distinct: bool = False,
-    ) -> Dict[str, int]:
+        sort_by: str = None
```
fridex: `sort_by` is of type `QuerySortTypeEnum`

thoth/storages/graph/postgres.py
```diff
@@ -3287,14 +3289,30 @@ def get_python_package_versions_all_count(
         if python_version is not None:
             query = query.filter(PythonPackageVersion.python_version == python_version)
 
+        if sort_by:
+            if sort_by == QuerySortTypeEnum.PACKAGE_VERSION.value:
+                query = query.order_by(func.count(
+                            PythonPackageVersion.package_version).desc()
```
fridex: This is not that simple - it's not just string sorting. We will need to sort based on result of the query (which makes pagination invalid).

thoth/storages/graph/postgres.py
```diff
@@ -3287,14 +3289,30 @@ def get_python_package_versions_all_count(
...
+                        )
+
+            elif sort_by == QuerySortTypeEnum.PACKAGE_NAME.value:
+                query = query.order_by(PythonPackageVersion.package_name)
+
+            else:
+                raise SortTypeQueryError(
+                        f"The sort key known in Thoth for queries are {[e.value for e in QuerySortTypeEnum]!r}"
```
fridex: As we add more sort criteria this will become ambiguous. You can raise with something like "unknown sort criteria for ... query"

thoth/storages/graph/postgres.py
```diff
@@ -3287,14 +3289,30 @@ def get_python_package_versions_all_count(
...
+                        )
+
+        groups_count = query.count()
```
fridex: Will this perform another query to the database?

pacospace: Yes it will

fridex: IMHO it would be better to reuse `_count` methods - for API simplicity and clearness.

pacospace: we don't have a `_count` for the groups created by a query using `group_by` method inside, we need a new one :)

pacospace:
PTAL @fridex :beers:

pacospace:
recheck

thoth/storages/graph/postgres.py
```diff
@@ -3318,13 +3341,25 @@ def get_python_package_versions_all_count(
             if python_version is not None:
                 query = query.filter(PythonPackageVersion.python_version == python_version)
 
+            if sort_by:
+                query = self._sort_by(
+                    query=query,
+                    query_name="get_python_package_versions_all_count",
```
fridex: I think this is not needed - no need to maintain it, we can read it from backtrace.

thoth/storages/graph/postgres.py
```diff
@@ -2992,6 +2994,26 @@ def get_software_environments_all(
...
             return [model.dict() for model in result]
 
+    def _sort_by(
+        self,
+        query: Query,
+        query_name: str,
+        sort_by: QuerySortTypeEnum
+    ) -> Query:
+        """Add sort to query, which is not executed."""
+        if sort_by == QuerySortTypeEnum.PACKAGE_NAME.value:
```
fridex: ```sort_by == QuerySortTypeEnum.PACKAGE_NAME```? `sort_by` is of type `QuerySortTypeEnum`

thoth/storages/graph/postgres.py
```diff
@@ -2992,6 +2994,25 @@ def get_software_environments_all(
...
+        self,
+        query: Query,
+        sort_by: QuerySortTypeEnum
+    ) -> Query:
+        """Add sort to query, which is not executed."""
+        if sort_by.value == QuerySortTypeEnum.PACKAGE_NAME.value:
+            query = query.order_by(PythonPackageVersion.package_name)
+
+        elif sort_by.value == QuerySortTypeEnum.PACKAGE_VERSION.value:
+            raise SortTypeQueryError("To be implemented.")  # TODO: To be implemented
```
fridex: This will be done after we receive results of the query.

thoth/storages/graph/models_base.py
```diff
@@ -100,3 +100,26 @@ def get_python_package_version_index_combinations(*, index_as_property: bool = T
             )
 
     return result
+
+
+class QueryResults:
```
fridex: I would probably name it `QueryResult` as it is a result of a query call (even though there are 2 queries under the hood) - and maybe move to another file - it does not look like it should belong to models file.

thoth/storages/graph/postgres.py
pacospace: So it become a bit tricky to generalize it :thinking:

thoth/storages/graph/enums.py
```diff
@@ -63,3 +63,10 @@ class MetadataDistutilsTypeEnum(Enum):
     # rarely used fields
     PROVIDED = "PROVIDED"
     OBSOLETE = "OBSOLETE"
+
+
+class QuerySortTypeEnum(Enum):
+    """Class for the requirements format."""
+
+    PACKAGE_NAME = "package_name"
+    PACKAGE_VERSION = "package_version"
```
fridex: In general, it's better to use int for enum values - comparing ints much less expensive operation than string comparison. Let's address it in a separate PR.

PR #1311, opened 2019-11-22
Simplified API functions

KPostOffice:
Reopen accidentally merged PR #1179 

Depends-On: #1464


thoth/storages/graph/postgres.py
```diff
@@ -5001,6 +5001,48 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
 
                 inspection_run.dependency_monkey_run_id = dependency_monkey_run.id
 
+    def get_python_package_required_symbols(
+        self, package_name: str, package_version: str, index_url: str
+    ) -> List[str]:
+        """Get required symbols for a Python package in a specified version."""
+        package_name = self.normalize_python_package_name(package_name)
+        package_version = self.normalize_python_package_version(package_version)
+
```
pacospace: Both queries require `with self._session_scope() as session:` with the new way to handle session safely.

thoth/storages/graph/postgres.py
```diff
@@ -5001,6 +5001,48 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+        query = self._session.query(PythonPackageVersion)
```
pacospace: Due to the new way to handle session we don't use  `self._session` but directly `session.query(...)` now.

thoth/storages/graph/postgres.py
```diff
@@ -5001,6 +5001,48 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+
+        query = (
+            query.filter(PythonPackageVersion.package_name == package_name)
+            .filter(PythonPackageVersion.package_version == package_version)
+            .join((HasArtifact, PythonPackageVersion.python_artifacts))
```
pacospace: It's not `PythonPackageVersion` but `PythonPackageVersionEntity` that has `python_artifacts` because PackageAnalyzer and PackageExtract cannot sync in `PythonPackageVersion` but in `PythonPackageVersionEntity` as constraint we have in the syncs.

thoth/storages/graph/postgres.py
```diff
@@ -5001,6 +5001,48 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
 
                 inspection_run.dependency_monkey_run_id = dependency_monkey_run.id
 
+    def get_python_package_required_symbols(
+        self, package_name: str, package_version: str, index_url: str
+    ) -> List[str]:
```
pacospace: Using `query.first()` you can have None indeed, therefore the output should be `Union[List[str], str]` because it can have None.

thoth/storages/graph/postgres.py
```diff
@@ -5001,6 +5001,48 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+
+
+    def get_image_symbols(
+        self,
+        environment_name: str,
+        os_name: str,
+        os_version: str,
+        cuda_version: str,
+        python_version: str
+    ) -> List[str]:
```
pacospace: Same comment as above, it should be  `Union[List[str], str]`

thoth/storages/graph/postgres.py
```diff
@@ -5001,6 +5001,48 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+        """Get symbols associated with a given image."""
+        query = (
+            self._session.query(PackageExtractRun)
+            .filter(PackageExtractRun.analysis_document_id == environment_name)
```
pacospace: What is `environment_name` in this case and why you filter that with the `analysis_document_id`? Maybe you mean `environment_type`? BUILDTIME and RUNTIME? In that case you should enforce the use of `EnvironmentTypeEnum`.

thoth/storages/graph/postgres.py
goern: and we convert None type to a 'None' string?

pacospace: +1 Indeed it can be `Union[List[str], None]` or better `Optional[List[str]]`. Thanks :)

goern:
@KPostOffice ping

thoth/storages/graph/postgres.py
```diff
@@ -4922,6 +4922,59 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+            else:
+                query = (
+                    session.query(PackageExtractRun)
+                    .filter(PackageExtractRun.os_name == os_name)
+                    .filter(PackageExtractRun.os_version == os_version)
+                    .filter(PackageExtractRun.python_version == python_version)
+                    .filter(PackageExtractRun.cuda_version == cuda_version)
+                    .with_entities(
+                        PackageExtractRun.versioned_symbols
+                    ))
```
pacospace: IMHO we could reduce the code splitting the pieces of the queries and adding optional only one filter :) WDYT?

```
query = (
    session.query(PackageExtractRun)
    .filter(PackageExtractRun.os_name == os_name)
    .filter(PackageExtractRun.os_version == os_version)
    .filter(PackageExtractRun.python_version == python_version)
)  

if cuda_version:
    query = query.filter(PackageExtractRun.cuda_version == cuda_version)

query = query.with_entities(
    PackageExtractRun.versioned_symbols
    )
```

KPostOffice: I like this, makes functionality more apparent and cuts down on the function size.

thoth/storages/graph/postgres.py
```diff
@@ -4922,6 +4922,53 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+        cuda_version: str,
+    ) -> Optional[List[str]]:
+        """Get symbols associated with a given image."""
+        with self._session_scope() as session:
+            query = (
+                session.query(PackageExtractRun)
+                .filter(PackageExtractRun.os_name == os_name)
+                .filter(PackageExtractRun.os_version == os_version)
+                .filter(PackageExtractRun.python_version == python_version)
+                .filter(PackageExtractRun.cuda_version == cuda_version)
```
pacospace: This line should not be here if `cuda_version` is optional :)

thoth/storages/graph/postgres.py
```diff
@@ -4679,6 +4679,53 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+        package_version = self.normalize_python_package_version(package_version)
+
+        with self._session_scope() as session:
+            query = session.query(PythonPackageVersion)
+
+            query = (
+                query.filter(PythonPackageVersion.package_name == package_name)
+                .filter(PythonPackageVersion.package_version == package_version)
+                .join((HasArtifact, PythonPackageVersionEntity.python_artifacts))
+                .join((PythonArtifact, HasArtifact.python_artifact))
```
fridex: I think you can use `join` without explicitly stating relation - sqlalchemy will use the default one:

```python
                .join(HasArtifact, PythonPackageVersionEntity)
                .join(PythonArtifact)
```

thoth/storages/graph/postgres.py
```diff
@@ -4679,6 +4679,53 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+                .with_entities(PythonArtifact.versioned_symbols)
+            )
+
+            return query.first()   # If query result is empty returns None
```
fridex: I think you want to return all of them, not just the first one.

thoth/storages/graph/postgres.py
```diff
@@ -4679,6 +4679,53 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+                .filter(PackageExtractRun.os_version == os_version)
+                .filter(PackageExtractRun.python_version == python_version)
+                .with_entities(
+                    PackageExtractRun.versioned_symbols
+                ))
+
+            if cuda_version:
+                query = query.filter(PackageExtractRun.cuda_version == cuda_version)
+
+            query = query.with_entities(PackageExtractRun.versioned_symbols)
```
fridex: Duplicate `with_entities`.

thoth/storages/graph/postgres.py
```diff
@@ -4679,6 +4679,53 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+
+            # If query result is empty returns None
+            return query.first()
```
fridex: Maybe `return query.all()` to return them all?

fridex:
@KPostOffice please rebase
@pacospace PTAL

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4671,77 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+                .join(
+                    PythonPackageIndex,
+                    PythonPackageIndex.id == PythonPackageVersionEntity.python_package_index_id
+                )
+                .outerjoin(
+                    VersionedSymbol,
+                    RequiresSymbol.versioned_symbol_id == VersionedSymbol.id
+                )
+                .filter(PythonPackageVersionEntity.package_name == package_name)
+                .filter(PythonPackageVersion.package_version == package_version)
```
pacospace: Should be `PythonPackageVersionEntity` I guess.

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4671,77 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+                .filter(PythonPackageIndex.url == index_url)
+                .with_entities(VersionedSymbol.symbol)
+            )
+
+            # Query returns list of single tuples
+            # NOTE: can be empty even if request is valid
+            return [i[0] for i in query.all()]
+
+    def get_image_symbols(
```
pacospace: what about `get_analyzed_image_symbols`? and so adjust the description please.

fridex: Based on the naming schema, should there be `_all` suffix?

pacospace: Correct, we get all now, not just one, thanks @fridex :)

thoth/storages/graph/postgres.py
fridex: I think the signature should define return type of `List[str]` - if there are no symbols, an empty list should be returned.

IMHO `first` returns [the very first result of the query](https://docs.sqlalchemy.org/en/13/orm/query.html). We want to retrieve all symbols, not just the first one.

thoth/storages/graph/postgres.py
KPostOffice: Thanks for the catch!

KPostOffice:
recheck

thoth/storages/graph/postgres.py
pacospace: But there is no `first` right now anymore, right?

pacospace: Here we need to add `_all` suffix as well instead :)

fridex: Yes, Kevin addressed my earlier comments already. The return value IMHO should be `List[str]`

pacospace: Yes, the comments above, where related to previous implementation when it was only one simbol and if not present was giving `None`.

thoth/storages/graph/models.py
```diff
@@ -162,6 +162,7 @@ class PythonPackageVersionEntity(Base, BaseExtension):
         Index(
             "python_package_version_entity_idx",
             "package_name",
+            "id",
```
fridex: Why `id` was added to this index?

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4671,80 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+        package_version = self.normalize_python_package_version(package_version)
+
+        with self._session_scope() as session:
+            query = (
+                session.query(PythonPackageVersionEntity)
+                .filter(PythonPackageVersionEntity.package_name == package_name)
+                .filter(PythonPackageVersionEntity.package_version == package_version)
+                .join(
+                    PythonPackageIndex,
+                    PythonPackageIndex.id == PythonPackageVersionEntity.python_package_index_id
```
fridex: Let sqlalchemy do its job here - no need to explicitly enumerate join conditions (here and bellow):

```python
                .join(PythonPackageIndex)
```

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4671,80 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+                    VersionedSymbol,
+                    RequiresSymbol.versioned_symbol_id == VersionedSymbol.id
+                )
+                .with_entities(VersionedSymbol.symbol)
+            )
+
+            # Query returns list of single tuples
+            # NOTE: can be empty even if request is valid
+            to_return = [i[0] for i in query.all()]
+            _LOGGER.debug("The package requires the following symbols: %r", str(to_return))
```
fridex: Please remove this statement, the string conversion will happen all the time making the debug statement non-optimal.

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4671,80 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+            return [i[0] for i in query.all()]
```
fridex: Unnecessary iteration over the result twice.

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4671,80 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+            query = (
+                session.query(PackageExtractRun)
+                .filter(PackageExtractRun.os_id == os_name)
+                .filter(PackageExtractRun.os_version_id == os_version)
+                .join(
+                    ExternalSoftwareEnvironment,
+                    PackageExtractRun.external_software_environment_id == ExternalSoftwareEnvironment.id
+                )
+                .filter(ExternalSoftwareEnvironment.cuda_version == cuda_version)
+                .filter(ExternalSoftwareEnvironment.python_version == python_version)
```
fridex: It will be good to add these filters conditionally I think. Otherwise, you will be matching to `None` values in tables if `python_version` and `cuda_version` will not be provided. (I guess that's not expected behaviour?)

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4671,80 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+                    VersionedSymbol,
+                    HasSymbol.versioned_symbol_id == VersionedSymbol.id
+                )
+                .with_entities(VersionedSymbol.symbol)
+            )
+
+            # Query returns list of single tuples (empty if bad request)
+            to_return = [i[0] for i in query.all()]
+            _LOGGER.debug("The image has the following symbols: %r", str(to_return))
+            return [i[0] for i in query.all()]
```
fridex: See my comments above.

thoth/storages/graph/postgres.py
KPostOffice: That was what I planned, if they are not provided I only wanted to match with rows that had `NULL` in those columns.  Otherwise, we might match multiple `cuda_versions` if we don't filter on it.

fridex: OK, that makes sense :+1: 

BTW please use `SoftwareEnvironment` - that's software environment we suggest users to use (e.g. our s2i-thoth container images). `ExternalSoftwareEnvironment` can be any environment that a user submitted for analysis and we do not have any control about it.

thoth/storages/graph/models.py
KPostOffice: It is used for joins

fridex: I think you want to create a new index on `id`. This way you have added a new key to the existing compound index.

KPostOffice: Okay, I'll make that change.

KPostOffice:
@fridex PTAL

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4675,56 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+            query = (
+                session.query(PackageExtractRun)
+                .filter(PackageExtractRun.os_id == os_name)
+                .filter(PackageExtractRun.os_version_id == os_version)
+                .join(SoftwareEnvironment)
+                .filter(SoftwareEnvironment.cuda_version == cuda_version)
+                .filter(SoftwareEnvironment.python_version == python_version)
+                .join(HasSymbol)
+                .join(VersionedSymbol)
+                .with_entities(VersionedSymbol.symbol)
```
fridex: It might be a good idea to add distinct here.

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4675,56 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+            query = (
+                session.query(PythonPackageVersionEntity)
+                .filter(PythonPackageVersionEntity.package_name == package_name)
+                .filter(PythonPackageVersionEntity.package_version == package_version)
+                .join(PythonPackageIndex)
+                .filter(PythonPackageIndex.url == index_url)
+                .join(HasArtifact)
+                .join(RequiresSymbol)
+                .join(VersionedSymbol)
+                .with_entities(VersionedSymbol.symbol)
```
fridex: Distinct as well.

thoth/storages/graph/postgres.py
```diff
@@ -4671,6 +4675,56 @@ def sync_dependency_monkey_result(self, document: dict) -> None:
...
+        *,
+        python_version: Optional[str] = None,
+        cuda_version: Optional[str] = None,
+    ) -> List[str]:
+        """Get symbols associated with a given image."""
+        with self._session_scope() as session:
+            query = (
+                session.query(PackageExtractRun)
+                .filter(PackageExtractRun.os_id == os_name)
+                .filter(PackageExtractRun.os_version_id == os_version)
```
fridex: `self.normalize_os_version`

thoth/storages/graph/models.py
```diff
@@ -262,6 +263,11 @@ class PackageExtractRun(Base, BaseExtension):
         "ExternalSoftwareEnvironment", back_populates="external_package_extract_runs"
     )
 
+    __table_args__ = (
+        Index("os_id_idx", "os_id"),
+        Index("os_version_id_idx", "os_id"),
```
fridex: Mind the index is duplicate - 2 indexes with defferent names introduced twice for the same column `os_id`.

To hit the index with query properly, there will be needed one more index with `os_id` and `os_version_id`.

thoth/storages/graph/postgres.py
fridex: BTW this will change the query semantics.

fridex:
recheck

PR #1328, opened 2019-11-25
Modify data type for metadata keys: Str -> Text

pacospace:
Fixes: https://github.com/thoth-station/storages/issues/1291
Fixes: https://github.com/thoth-station/storages/issues/1294
Fixes: https://github.com/thoth-station/storages/issues/1296
Fixes: https://github.com/thoth-station/storages/issues/1320
Fixes: https://github.com/thoth-station/storages/issues/1322
Fixes: https://github.com/thoth-station/storages/issues/1323


pacospace:
> What about using `Text` as its done in warehouse implementation? This way we will keep increasing forever
> 
> [#1323 (comment)](https://github.com/thoth-station/storages/issues/1323#issuecomment-557934189)

I was going to ask you that indeed :) Definetly better solution. Thanks!

pacospace:
Apparently also with the flag `compare_type = True` alembic is not able to find changes in datatype between String and Text (in character length yes). I will adjust the migration manually.

pacospace:
> Apparently also with the flag `compare_type = True` alembic is not able to find changes in datatype between String and Text. I will adjust the migration manually.

Please Test it as well, otherwise should we open issue in `alembic` project?

fridex:
> > Apparently also with the flag `compare_type = True` alembic is not able to find changes in datatype between String and Text. I will adjust the migration manually.
> 
> Please Test it as well, otherwise should we open issue in `alembic` project?

Awesome :+1: Will test it. And it might be worth to discuss this upstream.

pacospace:
recheck

pacospace:
recheck

Issue #1319, opened 2019-11-23
Optimized unsolved query


fridex:
**Describe the bug**

As our database grows, queries related to "unsolved" packages seem to not perform well. As we plan to ingest more data, we should optimize them.

**To Reproduce**
1. Go to stage deployment
2. Query for unsolved packages
3. See low performance of queries

**Expected behavior**

Queries should be more optimal.

**Additional context**

Currently, we use `NOT IN` which results in expensive queries, given the size of `PythonPackageVersionEntity` and `PythonPackageVersion` tables. We should switch to `NOT EXISTS`, which is a much more optimal solution for us. Here is a result of my experiment:

```python
from sqlalchemy import exists
from thoth.storages.graph.models import PythonPackageVersionEntity
from thoth.storages.graph.models import PythonPackageVersion
from thoth.storages.graph.models import PythonPackageIndex
from thoth.storages import GraphDatabase

graph = GraphDatabase()
graph.connect()
session = graph._sessionmaker()

query = (
    session.query(PythonPackageVersionEntity)
    .filter(
        PythonPackageVersionEntity.package_version.isnot(None),
        PythonPackageIndex.url.isnot(None),
        ~exists().where(PythonPackageVersionEntity.id == PythonPackageVersion.entity_id)
    )
    .join(PythonPackageIndex)
    .with_entities(
        PythonPackageVersionEntity.package_name,
        PythonPackageVersionEntity.package_version,
        PythonPackageIndex.url
    )
)

result = query.all()
session.close()
```

fridex:
```sql
SELECT python_package_version_entity.package_name AS python_package_version_entity_package_name, python_package_version_entity.package_version AS python_package_version_entity_package_version, python_package_index.url AS python_package_index_url
FROM python_package_version_entity JOIN python_package_index ON python_package_index.id = python_package_version_entity.python_package_index_id
WHERE python_package_version_entity.package_version IS NOT NULL AND python_package_index.url IS NOT NULL AND NOT (EXISTS (SELECT *
FROM python_package_version
WHERE python_package_version_entity.id = python_package_version.entity_id))
```

pacospace:
I tested suggested changes and I introduced a little change that seems to give also a little improvement in the performance.

```

Test

FUNCTION: get_unsolved_python_package_versions_count_all

current implementation

INPUTS: {'distinct': False, 'python_version': None, 'os_version': None, 'os_name': None, 'index_url': None, 'package_version': None, 'package_name': None}

[...]

PR #1329, opened 2019-11-25
Optimize queries

pacospace:
- Optimized solved with error queries 
- Optimized unsolved queries 
- Optimized analyzed queries
- Optimized analyzed with error queries
- Optimized unanalyzed queries
- Cleaner code
- Standardize query result output for python packages queries

Fixes: https://github.com/thoth-station/storages/issues/1319


thoth/storages/graph/postgres.py
```diff
@@ -959,31 +963,37 @@ def _construct_solved_query(
...
 
         if package_version is not None:
             package_version = self.normalize_python_package_version(package_version)
-            query = query.filter(PythonPackageVersion.package_version == package_version)
+            query = query.filter(PythonPackageVersionEntity.package_version == package_version)
 
-        if os_name is not None:
-            query = query.filter(PythonPackageVersion.os_name == os_name)
+        conditions = []
+        conditions.append(PythonPackageVersionEntity.id == PythonPackageVersion.entity_id)
```
fridex:
```python
conditions = [PythonPackageVersionEntity.id == PythonPackageVersion.entity_id]
```

thoth/storages/graph/query_result_base.py
```diff
@@ -27,8 +29,8 @@ class QueryResult:
 
     def __init__(
         self,
-        result: dict,
-        count: int
+        result: Union[List, Dict],
```
fridex:
```
        result: Union[List, Dict[str, Any]],
```

thoth/storages/graph/query_result_base.py
```diff
@@ -43,3 +45,53 @@ def result(self):
     def count(self):
         """Query count result."""
         return self._count
+
+    def count_per_package(self) -> Dict[Tuple[str, str, str], int]:
```
fridex: By adding these methods, the class becomes specific to Python packages - maybe rename to `PythonQueryResult` or something similar?

fridex: Also, I'm not sure whether this approach is safe.

pacospace: I thought would give us clean code, also because more functions use similar output, so in this class we refactor the output of the queries to be in the way we need, but all queries give result using sqalchemy function, e.g. count() or all(). Alternatively I can take everything back :)

pacospace: Do you mean we have higher risk to break more methods at once?

pacospace:
recheck

pacospace:
recheck

goern:
Do we have metrics so that we can evaluate the performance improvement of this change?

pacospace:
> Do we have metrics so that we can evaluate the performance improvement of this change?

Some of the queries were not running because of the size of the database, therefore we had to optimize them. Now for the current size of the database, they all run and pretty fast :)

thoth/storages/graph/query_result_base.py
fridex: With this change we will need to make sure `PythonQueryResult` uses the structure of results so users of `PythonQueryResult` can call methods safely. Is this the case? - All methods which return `PythonQueryResult` object can safely call these methods?

goern:
shall we merge this and do another release? @fridex @pacospace

fridex:
> shall we merge this and do another release? @fridex @pacospace

Let's [clarify usage based on my comment](https://github.com/thoth-station/storages/pull/1329#discussion_r351144108), we can do that then :+1:

pacospace:
> > shall we merge this and do another release? @fridex @pacospace
> 
> Let's [clarify usage based on my comment](https://github.com/thoth-station/storages/pull/1329#discussion_r351144108), we can do that then

Updated :) Thanks :beers: Should we also have this: https://github.com/thoth-station/storages/pull/1328 before releasing?

fridex:
> > > shall we merge this and do another release? @fridex @pacospace
> > 
> > 
> > Let's [clarify usage based on my comment](https://github.com/thoth-station/storages/pull/1329#discussion_r351144108), we can do that then
> 
> Updated :) Thanks  Should we also have this: #1328 before releasing?

Thanks! :+1: Yes :beers:

goern:
recheck

pacospace:
recheck

goern:
regate

PR #1332, opened 2019-11-25
Fix Key Error on Operating System

bissenbay:
Fixes: https://github.com/thoth-station/storages/issues/1330

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -4284,8 +4284,8 @@ class GraphDatabase(SQLBase):
         environment_type = environment_type.upper()
         origin = document["metadata"]["arguments"]["thoth-package-extract"]["metadata"].get("origin")
         environment_name = document["metadata"]["arguments"]["extract-image"]["image"]
-        os_name = document["result"]["operating-system"]["name"]
-        os_version = document["result"]["operating-system"]["version_id"]
+        os_name = document["result"].get("operating-system", {}).get("name")
+        os_version = document["result"].get("operating-system", {}).get("version_id")
 
         # Check if it comes from a User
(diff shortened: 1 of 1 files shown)


bissenbay:
@fridex Here some installation failure happening which I'm facing first time.

fridex:
recheck

thoth/storages/graph/postgres.py
```diff
@@ -4284,8 +4284,8 @@ def sync_analysis_result(self, document: dict) -> None:
         environment_type = environment_type.upper()
         origin = document["metadata"]["arguments"]["thoth-package-extract"]["metadata"].get("origin")
         environment_name = document["metadata"]["arguments"]["extract-image"]["image"]
-        os_name = document["result"]["operating-system"]["name"]
-        os_version = document["result"]["operating-system"]["version_id"]
+        os_name = document["result"].get("operating-system", {}).get("name")
+        os_version = document["result"].get("operating-system", {}).get("version_id")
```
fridex: The exception looks suspicious, these values should be always present:

https://github.com/thoth-station/package-extract/blob/54080efc724c0b353c85024f0901a6485a43b040/thoth/package_extract/image.py#L543-L545

fridex: @bissenbay could you verify package-extract run (how was it triggered and what results it produced). package-extract might be also in an old version in deployments.

bissenbay:
@fridex This is my fault, I was analyzing this image `quay.io/thoth-station/s2i-thoth-fedora30-py37:v0.5.0` and only was interested in `python-interpreters` and commented the rest to speed up the process of getting results faster. I guess, this can be closed and it is not relevant.

PR #1340, opened 2019-11-27
Gather document id from document_id field

fridex:


goern:
recheck

fridex:
recheck

goern:
recheck

PR #1341, opened 2019-11-27
Fix Issue #1308 not iterable

bissenbay:
Fixes: https://github.com/thoth-station/storages/issues/1308


thoth/storages/graph/postgres.py
```diff
@@ -4889,43 +4889,45 @@ def sync_adviser_result(self, document: dict) -> None:
             )
 
             # Output stacks - advised stacks
-            for idx, result in enumerate(document["result"]["report"]):
```
fridex: Is this the following change?
```python
            for idx, result in enumerate(document["result"].get("report", [])):	
```

goern:
recheck

thoth/storages/graph/postgres.py
bissenbay: Thanks @fridex , this is way better

goern:
recheck

PR #1366, opened 2019-12-05
Add aggregated_at column to CVE

bissenbay:


thoth/storages/graph/postgres.py
```diff
@@ -3852,7 +3853,12 @@ def create_python_cve_record(
         package_version = self.normalize_python_package_version(package_version)
         with self._session_scope() as session, session.begin(subtransactions=True):
             cve, _ = CVE.get_or_create(
-                session, cve_id=record_id, version_range=version_range, advisory=advisory, cve_name=cve
+                session,
+                cve_id=record_id,
+                version_range=version_range,
+                advisory=advisory,
+                cve_name=cve,
+                aggregated_at=datetime.now(),
```
bissenbay: @fridex I wanted to add `datetime.now()` on this line https://github.com/bissenbay/cve-update-job/blob/master/app.py#L94. However, as we did not want to capture the download time I chose this place. However, the only concern I have here is that this will create duplicates of CVE records with different datetime.

fridex: Yes, that's true. :+1: 

Let's check then whether the database always states uniquely `cve_id` or `cve_name` and try to query based on it.

thoth/storages/graph/postgres.py
```diff
@@ -3852,8 +3853,22 @@ def create_python_cve_record(
...
             cve, _ = CVE.get_or_create(
-                session, cve_id=record_id, version_range=version_range, advisory=advisory, cve_name=cve
+                session,
+                cve_id=record_id,
+                version_range=version_range,
+                advisory=advisory,
+                cve_name=cve,
+                # aggregated_at=datetime.now(),
             )
+            if cve.aggregated_at is None:
```
bissenbay: @fridex WDYT about this approach? The idea is to get or create a CVE instance based on all parameters but `aggregated_at`. As we indicate `nullable=True`, the instance should have `None` if it is newly created. Once we detect a new row data, then we could set `aggregated_at` attribute of the object to `datetime.now()` and next you could just find that row based on `cve.id` which is primary key I guess, then update that record.

fridex: The issue is in the `get_or_create` which does exact matching. Each time it will go through the records in the database and will find out there is no record that would have:

```python
                cve_id=record_id,
                version_range=version_range,
                advisory=advisory,
                cve_name=cve,
                aggregated_at=None
```

... so it will be created. The code below will set `aggregated_at` and insert such record. In the next round, there will be performed exact matching again (with `aggregated_at` set to `None`) so there will be no record created (again). You can try to replicate this if you run the CVE syncing logic multiple times.

thoth/storages/graph/postgres.py
bissenbay: Most of the times `cve_name` is just empty string, but `cve_id` seems to be unique, so what about if we use both `cve_id` and `cve_name` instead of just `cve_id`?

fridex: If `cve_id` is unique, I think we should be just fine to use it not to complicate things.

thoth/storages/graph/postgres.py
bissenbay: Should I really use `get_or_create`? What if I create a static method something like `get` in `BaseExtension` class which would check if the record exists on specific attribute values?

fridex: You can directly querying logic as provided by sqlalchemy methods. [Here for example, we query package extract runs](https://github.com/thoth-station/storages/blob/0e0ee17a00f8aa0bc726e477344ee107b5e22090/thoth/storages/graph/postgres.py#L344-L352). Docs are available here - https://docs.sqlalchemy.org/en/13/orm/query.html

thoth/storages/graph/postgres.py
```diff
@@ -3851,9 +3852,17 @@ def create_python_cve_record(
         package_name = self.normalize_python_package_name(package_name)
         package_version = self.normalize_python_package_version(package_version)
         with self._session_scope() as session, session.begin(subtransactions=True):
-            cve, _ = CVE.get_or_create(
-                session, cve_id=record_id, version_range=version_range, advisory=advisory, cve_name=cve
-            )
+            cve = session.query(CVE).filter(CVE.cve_id == record_id).first()
```
bissenbay: @fridex I just left `with_entities` out, so guessing that way all will request parameters like in SQL `*`.

fridex: To optimize the query, you can use exists - that way the database will immediately return if matching result is found and will not return all the data (serialization/deserialization + data transfer overhead):

```python
session.query(exists().where(CVE.id==record_id))
```

It will also make sense to make `cve_id` not nullable and unique (indexed for better performance).

bissenbay: Thanks for the advise. But looks like `cve = session.query(exists().where(CVE.id==record_id))` just constructs the query and if I add `scalar()` to the end of the query like `cve = session.query(exists().where(CVE.id==record_id)).scalar()`, then it returns a boolean whether the record exists or not.

bissenbay: @fridex , should I stick to the first one which is `cve = session.query(CVE).filter(CVE.cve_id == record_id).first()` instead of `session.query(exists().where(CVE.id==record_id))`?

fridex: Use the `exists` one, it has benefits stated above.

bissenbay: @fridex I'm sorry, but I really want to understand. If we go with `exists` approach, then as I said it return True or False, right. Should I like, first find out whether that record exists, then execute another query to really get that record because we should return CVE object?
That's why for me it looks like 2 queries of which 1 is checking for existence and 2 is to really getting that data. However, I was suggesting instead to issue only one query where it tries to get the data immediately, so if we see that `None` is returned, then we can issue `CVE.get_or_create()` which will create that data, otherwise if something returned, we will return it?

fridex: The most used scenario is when we already have the given CVE record (as we run this component CVE update job as a cron job for the same set of CVEs, that can get updated over time). So in most cases, we will hit "CVE is already present in the database" - that's why I'm suggesting that simple `exists` check that has minimal overhead.

If we do not have the given CVE with `cve_id` in the database, we can create it (not with `get_or_create` as we already know its not in the database).

thoth/storages/graph/postgres.py
```diff
@@ -3846,26 +3847,35 @@ def create_python_cve_record(
         version_range: str,
         advisory: str,
         cve: str = None,
-    ) -> Tuple[CVE, bool]:
```
bissenbay: @fridex at looked at https://github.com/thoth-station/cve-update-job/blob/master/app.py#L87 and it looks for me that we don't need to return CVE at all because we are not using it in `cve-update-job`. Rather, what we are interested is whether we have that cve already in the database that's why WDTY if we return True only if the record exists, otherwise False and we going to log about it accordingly. PTAL at https://github.com/thoth-station/cve-update-job/pull/251/?

fridex: The approach sounds reasonable :+1:

thoth/storages/graph/postgres.py
```diff
@@ -3846,26 +3847,35 @@ def create_python_cve_record(
...
-                package_version=package_version,
-                python_package_index_id=index.id,
-            )
-            _, existed = HasVulnerability.get_or_create(
-                session, cve_id=cve.id, python_package_version_entity_id=entity.id
-            )
+            if session.query(exists().where(CVE.cve_id == record_id)).scalar():
+                return True
+            else:
+                cve, _ = CVE.get_or_create(
```
fridex: You don't need to use `get_or_create` method here - it performs one check if the given CVE exists - you have already done it before. You can directly create the `CVE` entry and commit it to the database.

thoth/storages/graph/models.py
```diff
@@ -369,8 +369,9 @@ class CVE(Base, BaseExtension):
 
     advisory = Column(String(16384), nullable=True)
     cve_name = Column(String(256), nullable=True)
-    cve_id = Column(String(256), nullable=True)
+    cve_id = Column(String(256), nullable=False)
```
fridex: Please make this column unique.

thoth/storages/graph/postgres.py
```diff
@@ -3846,26 +3847,48 @@ def create_python_cve_record(
...
+                    "cve_id": record_id,
+                    "version_range": version_range,
+                    "aggregated_at": datetime.now(),
+                }
+                try:
+                    session.begin_nested()
+                    cve_instance = CVE(**kwargs)
+                    session.add(cve_instance)
+                    session.commit()
+                except IntegrityError as exc:
```
fridex: This is already handled by the context manager. You can simply do:

```python
                    cve_instance = CVE(
                        # params...
                    )
                    session.add(cve_instance)
```

The context manager will make sure everything is committed or rolled back on any failure.

bissenbay: Wow, thanks, I was figuring that out, :smile:

thoth/storages/graph/postgres.py
```diff
@@ -3846,26 +3847,35 @@ def create_python_cve_record(
...
+
+                index = self._get_or_create_python_package_index(session, index_url, only_if_enabled=False)
+                entity, _ = PythonPackageVersionEntity.get_or_create(
+                    session,
+                    package_name=package_name,
+                    package_version=package_version,
+                    python_package_index_id=index.id,
+                )
+                HasVulnerability.get_or_create(
+                    session, cve_id=cve_instance.id, python_package_version_entity_id=entity.id
```
bissenbay: @fridex My only concern is here whether that object `cve_instance` would have `id` parameter which is autoincremented, primary key, otherwise LGTM?

fridex: Please do not make primary key out of it (it's better to have integer primary keys) - make it unique and let PosgreSQL create an index for it.

You can try and test it :)

PS: You can also try `cve=cve_instance`. Let me know which version works for you.

bissenbay: Frido there are two things we have in `CVE`. First is `id` which is https://github.com/thoth-station/storages/blob/master/thoth/storages/graph/models.py#L368 and second `cve_id` which is https://github.com/thoth-station/storages/blob/master/thoth/storages/graph/models.py#L372. The first is `primary key, autoincremented` and the second one is `unique`.
Here, we are accessing `id`, not `cve_id`. Is my question clear now, I'm sorry for not explaining in the beginning :smile:

bissenbay: So, the is that when we add a new record to the table, then that `id` is generated, right? Now, we want to access it, I'm not sure whether we can access it like in this way.

fridex: Ah, right! Sorry, I overlooked it :)

Try with:

```
HasVulnerability.get_or_create(
                    session, cve_id=cve_instance.id, python_package_version_entity_id=entity.id
)
```

and with:

```
HasVulnerability.get_or_create(
                    session, cve=cve_instance, python_package_version_entity_id=entity.id
)
```

And ignore my comment regarding primary key - friday, too late... My brain is thinking about :beers: :smile:

bissenbay: @fridex Thanks, I checked both of them and this `cve_id=cve_instance.id,` works.

thoth/storages/graph/postgres.py
```diff
@@ -3846,26 +3847,35 @@ def create_python_cve_record(
...
-            )
+            if session.query(exists().where(CVE.cve_id == record_id)).scalar():
+                return True
+            else:
+                cve_instance = CVE(
+                    advisory=advisory,
+                    cve_name=cve,
+                    cve_id=record_id,
+                    version_range=version_range,
+                    aggregated_at=datetime.now(),
```
fridex: ``datetime.utcnow()`` and this can go in :wink:

PR #1386, opened 2019-12-12
Sync cuda version

bissenbay:
Fixes: https://github.com/thoth-station/storages/issues/934

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -154,6 +154,7 @@ from ..exceptions import DatabaseNotInitialized
 from ..exceptions import SolverNameParseError
 from ..exceptions import DistutilsKeyNotKnown
 from ..exceptions import SortTypeQueryError
+from ..exceptions import CudaVersionDoesNotMatch
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -4163,6 +4163,9 @@ def sync_analysis_result(self, document: dict) -> None:
         environment_name = document["metadata"]["arguments"]["extract-image"]["image"]
         os_name = document["result"]["operating-system"]["name"]
         os_version = document["result"]["operating-system"]["version_id"]
+        cuda_version_txt = document["result"]["cuda-version"]["/usr/local/cuda/version.txt"]
+        cuda_version_nvcc = document["result"]["cuda-version"]["nvcc_version"]
+        cuda_version = cuda_version_txt if cuda_version_txt == cuda_version_nvcc else None
```
fridex: Let's raise an exception here so we get notified about this fact and the sync will be aborted - at least now in the initial implementation.

bissenbay: Thanks!

thoth/storages/graph/postgres.py
```diff
@@ -4163,6 +4164,9 @@ def sync_analysis_result(self, document: dict) -> None:
         environment_name = document["metadata"]["arguments"]["extract-image"]["image"]
         os_name = document["result"]["operating-system"]["name"]
         os_version = document["result"]["operating-system"]["version_id"]
+        cuda_version = document["result"]["cuda-version"]["/usr/local/cuda/version.txt"]
+        if cuda_version != document["result"]["cuda-version"]["nvcc_version"]:
+            raise CudaVersionDoesNotMatch
```
fridex: Let's propagate this fact so that we see it in reporting tools:

```python
raise CudaVersionDoesNotMatch(f"Cuda version detect by nvcc {cuda_version!r} is different from the one found in /usr/local/cuda/version.txt {document['result']['cuda-version']['nvcc_version']!r}")
```

bissenbay: Thanks, done!

PR #1387, opened 2019-12-12
Introduce PyBench PI table and adjust sync logic for inspection

pacospace:


thoth/storages/graph/models_performance.py
```diff
@@ -37,8 +37,8 @@
 class PerformanceIndicatorBase:
     """A base class for implementing performance indicators."""
 
-    # ML framework used for the performance indicator.
-    framework = Column(String(256), nullable=True)
+    # library used for the performance indicator.
+    library = Column(String(256), nullable=True)
```
fridex: Not sure with `library` naming here - we have also Python specific performance indicators.

thoth/storages/graph/models_performance.py
```diff
@@ -80,9 +80,9 @@ def create_from_report(
     ) -> "PerformanceIndicatorBase":
         """Create performance indicator record together with related observed performance edge based on inspection."""
         # Place core parts of the base class into the model.
-        framework = inspection_document["job_log"]["stdout"].get("framework")
-        if not framework:
-            _LOGGER.warning("No machine learning framework specified in performance indicator %r", cls.__name__)
+        library = inspection_document["job_log"]["stdout"].get("library")
+        if not library:
+            _LOGGER.warning("No library specified in performance indicator %r", cls.__name__)
```
fridex: Do we want to sync such results? I'm wondering about their usefulness.

pacospace: This is required to distinguish between `tensorflow` or `pytorch`, now we have `pybench`, these are python modules at the end. So I thought library might be good.

thoth/storages/graph/models_performance.py
pacospace: What about `module`?

thoth/storages/graph/postgres.py
```diff
@@ -3744,31 +3749,22 @@ def _create_python_software_stack(
...
-                    software_environment=software_environment,
-                    sync_only_entity=sync_only_entity
-                )
-                for python_package_version_entity in python_package_versions_entities:
-                    ExternalPythonRequirementsLock.get_or_create(
-                        session,
-                        python_software_stack_id=software_stack.id,
-                        python_package_version_entity_id=python_package_version_entity.id,
-                    )
+                python_requirement_lock = ExternalPythonRequirementsLock
```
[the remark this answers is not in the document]
pacospace: should we change it to `external`?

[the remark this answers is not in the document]
fridex: Yes please, it should be more readable in sources.

thoth/storages/graph/models_performance.py
fridex: I was pointing to a case when it is not set. It might be good idea to refuse results that do not have `library` (in this case) set - we will not use them anyway as they are not bound to any library.

If we will just perform performance indicator tests, then I think we shouldn't trigger graph sync (this logic would be in Argo workflows).

thoth/storages/graph/models_performance.py
fridex: I think that's better. What about `component`?

pacospace: It is also good, I will change it in PI as well :)

thoth/storages/graph/postgres.py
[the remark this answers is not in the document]
pacospace: `sync_only_entity` is distributed across several functions down to create PPV which creates PPV and PPVE or only PPVE.
```
            python_package_versions = self._create_python_packages_pipfile(
                session,
                requirements_lock,
                software_environment=software_environment,
                sync_only_entity=is_external
            )
```

thoth/storages/graph/models_performance.py
pacospace: But the graph-sync would tell us if that key is missing, we can make it mandatory in the schema (nullable=False) so if that is not provided, it fails. So we don't need:

```
library = inspection_document["job_log"]["stdout"].get("library")
        if not library:
            _LOGGER.warning("No library specified in performance indicator %r", cls.__name__)
```

thoth/storages/graph/postgres.py
```diff
@@ -4104,7 +4090,7 @@ def _python_sync_analysis_result(
                 os_version=software_environment.os_version,
                 python_version=software_environment.python_version,
                 index_url=None,
-                sync_only_entity=True
+                is_external=True
```
fridex: This depends on the actual image analysis run - if it is external or analysis of an image that we plan to use in recommendations- there should be a metadata field propagated - https://github.com/thoth-station/user-api/blob/967285e46893f68d79b7f4ae5da81ed6c50b0e9d/thoth/user_api/api_v1.py#L82

thoth/storages/graph/postgres.py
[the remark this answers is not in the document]
fridex: Hm... maybe it will be better to keep both - `sync_only_entity` and `is_external`. WDYT?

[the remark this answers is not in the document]
fridex: All LGTM :+1: except for this - do we want to keep both or just one?

I think the main part where it makes sense is image analysis result syncs that require both flags, if I read the code correctly.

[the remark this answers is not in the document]
pacospace: We can keep both!

pacospace:
Code climate change should be ignored in this case.

fridex:
@pacospace please rebase.

pacospace:
> @pacospace please rebase.

I rebased, why it is not updating?

thoth/storages/graph/postgres.py
```diff
@@ -3565,6 +3571,7 @@ def _create_python_packages_pipfile(
                         os_version=os_version,
                         python_version=python_version,
                         index_url=package.index.url if package.index else None,
+                        metadata=python_package_metadata,
```
fridex: The called function uses `metadata` just to pick `id`. Can it accept also `metadata_id` and assign `python_package_metadata_id` accordingly? This way we will not issue the query introduced on line 3561:

https://github.com/thoth-station/storages/pull/1387/files#diff-42370be78220d49029e40a3fbfc0f09fR3561

thoth/storages/graph/postgres.py
```diff
@@ -4104,7 +4092,7 @@ def _python_sync_analysis_result(
                 os_version=software_environment.os_version,
                 python_version=software_environment.python_version,
                 index_url=None,
-                sync_only_entity=True
+                sync_only_entity=document["metadata"]["arguments"]["thoth-package-extract"]["metadata"]["is_external"]
```
fridex: `is_external`?

pacospace: mmm this is `_create_python_package_version` it does not have `is_external`

thoth/storages/graph/postgres.py
pacospace: Right!! Thanks!!

PR #1414, opened 2020-01-02
Optimize marker evaluation result query for adviser

fridex:


fridex:
@pacospace ping? so we can issue a new release :package:

PR #1443, opened 2020-01-07
Introduce software environment specific queries

fridex:


thoth/storages/graph/postgres.py
```diff
@@ -573,6 +573,35 @@ def python_package_exists(self, package_name: str) -> bool:
...
+                    PythonPackageVersion.os_name == os_name,
+                    PythonPackageVersion.os_version == os_version,
+                    PythonPackageVersion.python_version == python_version,
+                )
+                .exists()
+            ).scalar()
+
+            return result
+
+    def get_solved_software_environment_configurations(self) -> List[Dict[str, str]]:
```
pacospace: I don't think it follow the query convention. `get_<objects>_all()` where objects are entities from the schema table (e.g. PythonPackageVersion). In this case, we are taking property of PPV, not an entity from the schema, so it would be confusing maybe. WDYT? What about `get_solved_python_package_versions_software_environment_all`? (even if solved can be implicit because PPV are solved by definition)

fridex: Awesome! Thanks for the review. Adjusted in 68c4482

Issue #1450, opened 2020-01-07
'thoth.storages.graph.models.ExternalPythonRequirementsLock'>' has no property 'python_package_version_id'


fridex:
```
---> Running application from Python script (app.py) ...
--
  | 2020-01-07 18:23:50,989 [1] INFO     root:224: Setting up logging to a Sentry instance 'sentry.io/1298083', environment 'thoth-psi-stage' and integrations ['SqlalchemyIntegration']
  | 2020-01-07 18:23:50,994 [1] INFO     root:261: Logging to rsyslog endpoint is turned off
  | 2020-01-07 18:23:50,995 [1] INFO     thoth.graph_sync_job:238: graph-sync-job v0.6.0+thoth_storage.0.20.5+thoth_common.0.9.22 starting...
  | 2020-01-07 18:23:51,059 [1] INFO     alembic.runtime.migration:154: Context impl PostgresqlImpl.
  | 2020-01-07 18:23:51,060 [1] INFO     alembic.runtime.migration:161: Will assume transactional DDL.
  | 2020-01-07 18:23:51,496 [1] INFO     thoth.storages.sync:82: Syncing adviser document from 'https://s3.upshift.redhat.com/'  with id 'adviser-401a385e' to graph
  | 2020-01-07 18:23:51,535 [1] WARNING  thoth.storages.graph.postgres:4748: No origin stated in the adviser result 'adviser-401a385e'
  | 2020-01-07 18:23:51,535 [1] WARNING  thoth.storages.graph.postgres:4751: No s2i flag stated in the adviser result 'adviser-401a385e'
  | 2020-01-07 18:23:51,693 [1] CRITICAL root:101: Traceback (most recent call last):
  | File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/orm/base.py", line 399, in _entity_descriptor
  | return getattr(entity, key)
  | AttributeError: type object 'ExternalPythonRequirementsLock' has no attribute 'python_package_version_id'
  | During handling of the above exception, another exception occurred:
  | Traceback (most recent call last):
  | File "app.py", line 239, in <module>
  | cli()
  | File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 764, in __call__
  | return self.main(*args, **kwargs)
  | File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 717, in main
  | rv = self.invoke(ctx)
[...]

PR #1451, opened 2020-01-07
Fix syncing external software environments coming from adviser

fridex:
Fixes: #1450


fridex:
@pacospace the code is starting to be... ehm... we should refactor some of the parts.

fridex:
Tested locally, works. I will merge this and issue a new release to make pieces moving again in stage env. Feel free to raise any comments.

pacospace:
What do you have in mind? 🙂

fridex:
> What do you have in mind?

Fix typing and some of the methods could be refactored a bit. But nothing big we would need immediately.

Issue #1437, opened 2020-01-06
Remove id from relational tables that do not require it


fridex:
**Is your feature request related to a problem? Please describe.**

Having additional unused columns is not a good practice as they create additional overhead when storing or loading data - let's remove `id` column from `depends_on` table and other relation tables that do not require it.

PR #1457, opened 2020-01-08
Drop id columns on relation tables

fridex:
... to reduce footprint, cache size and limits for large tables (e.g. 
depends_on) which could easily overflow given the size of our database and
with the current id datatype used.

Fixes: #1437

Depends-On: #1456


fridex:
I've run a few tests not to corrupt our database. @pacospace could you please double-check that everything is ok? In the worst case, we still have backups. Thanks! :+1:

pacospace:
> I've run a few tests not to corrupt our database. @pacospace could you please double-check that everything is ok? In the worst case, we still have backups. Thanks!

I checked for queries that rely on those properties and I identified:
- `has_python_solver_error`
- `get_relation_table_count`

fridex:
> There are two queries affected

I'm not sure if I follow. Does it mean these two queries failed/

pacospace:
> > There are two queries affected
> 
> I'm not sure if I follow. Does it mean these two queries failed/

Yes if we remove the `id` from relation tables, those two queries are affected.

fridex:
> > > There are two queries affected
> > 
> > 
> > I'm not sure if I follow. Does it mean these two queries failed/
> 
> Yes if we remove the `id` from relation tables, those two queries are affected.

66efe72 addresses `get_relation_table_count`

fridex:
`1b7dab5` addresses `has_python_solver_error`. I think we don't need explicit order here.

fridex:
> `1b7dab5` addresses `has_python_solver_error`. I think we don't need explicit order here.

Hm, actually we do to take into account the most recent solver. Adjusted in 0078b5c. PTAL and thanks for testing!

fridex:
To make sure we do not corrupt/lose any data, I've created a backup in stage environment. The backup is available on Ceph - object key prefix is `data/thoth/thoth-psi-stage/graph-backup/pg_dump-20-01-08-16-57-52`.

PR #1495, opened 2020-01-13
Fix query for enabled index

fridex:


fridex:
`first()` returns a tuple in this case, we want to return boolean:

```python
>>> from thoth.storages import GraphDatabase
>>> graph = GraphDatabase()
>>> graph.connect()
>>> graph.is_python_package_index_enabled("https://pypi.python.org/simple")
(False,)
```

fridex:
I'll merge this and issue a new update so that we have updated components tomorrow.

PR #1554, opened 2020-01-28
Sync missing packages if adviser failed due to unknown dependencies

fridex:


fridex:
> Isn't Adviser considering only packages in PPV? How it can identify unknown dependencies if by definition in PPV we know everything of the packages?

Adviser states what dependencies were not resolved yet in the output report (the document that is being synced). Subsequently, graph sync creates entities in the database so that package-releases-job can discover them on Python indexes  (as these packages are considered as "seen") and graph-refresh-job can schedule solvers for the discovered ones.

fridex:
recheck

fridex:
regate

fridex:
regate

PR #1569, opened 2020-02-05
No need to query for package extract run - software environment can be directly used

fridex:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


fridex:
recheclk

fridex:
recheck

Issue #1571, opened 2020-02-05
Wrong operating system name synced from container image analysis


fridex:
**Describe the bug**

We sync the wrong operating system detect - instead of syncing OS id, we sync operating system name which can be "`Red Hat Enterprise Linux`" (instead of OS id which is "`rhel`").

**To Reproduce**
1. Run container image analysis
2. Explore data synced into our knowledge base
3. See the wrong operating system name

**Additional context**

PR #1572, opened 2020-02-05
Fix OS name synced in container image analysis

fridex:
Fixes: https://github.com/thoth-station/storages/issues/1571

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3953,7 +3953,7 @@ class GraphDatabase(SQLBase):
         environment_type = environment_type.upper()
         origin = document["metadata"]["arguments"]["thoth-package-extract"]["metadata"].get("origin")
         environment_name = document["metadata"]["arguments"]["extract-image"]["image"]
-        os_name = document["result"]["operating-system"]["name"]
+        os_name = document["result"]["operating-system"]["id"]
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -3953,7 +3953,7 @@ def sync_analysis_result(self, document: dict) -> None:
         environment_type = environment_type.upper()
         origin = document["metadata"]["arguments"]["thoth-package-extract"]["metadata"].get("origin")
         environment_name = document["metadata"]["arguments"]["extract-image"]["image"]
-        os_name = document["result"]["operating-system"]["name"]
```
pacospace: Didn't we wanted to have the same structure for all components?, like adviser, inspection, etc which uses `operating_system` and `name`. the `id` is given from `os_release`

fridex: `os_release` comes from generic analyzer metadata - package-extract has "os_info" handler for this

https://github.com/thoth-station/package-extract/blob/93b2800066114750700e4dcd9c5f167b246c6064/thoth/package_extract/image.py#L606-L607

Under the hood, they use similar code now. I wanted to sync this from "result", not from package-extract "metadata".

pacospace: Thanks for the clarification, then LGTM!! :100:

PR #1577, opened 2020-02-05
Address issue #1573

### LINKED ISSUE #1573: Check found Python interpreter when obtaining package symbols
**Describe the bug**

The current logic behind the container image symbols query takes into account Python version. Unfortunatelly, this value is always `null` in our database:

https://github.com/thoth-station/storages/blob/937f65356c5f6f022e0da87be3d67c1c07c45f52/thoth/storages/graph/postgres.py#L3985-L3987

**To Reproduce**
1. Run container image analysis
2. Wait container image analysis results to be synced
3. Check synced results and see `python_version` column
4. See `null` value

**Expected behavior**

We should be able to query for Python version when obtaining package symbols in the given software environment.

**Additional context**

As the symbols query on adviser side expects our software environments (not the ones being analyzed from external sources), we can explicitly provide Python version based on container image name we analyze (this is a special case for "SoftwareEnvironment", cannot be generalized for "ExternalSoftwareEnvironment"). This can be parsed from container image name as we have a naming convention for them, such as `quay.io/thoth-station/s2i-thoth-ubi8-py36` (`3.6` in this case).

If the container image name fails to be parsed, the sync should not happen (the transaction will be rolled back).

KPostOffice:
fixes #1573

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

Parse package version from image name when it comes from an internal source

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -17,6 +17,7 @@
 
 """An SQL database for storing Thoth data."""
 
+import re
 import logging
 import json
 import os
@@ -3981,6 +3982,8 @@ class GraphDatabase(SQLBase):
                 sw_class = SoftwareEnvironment
                 python_version = image_name.rsplit("-", maxsplit=1)[1]   # pyXX
                 python_version = python_version[2:3] + "." + python_version[3:]   # take first digit and put . after it
+                if not re.match(r"^\d\.\d*$", python_version):
+                    raise ValueError("Python Version does not match pattern")
 
             software_environment, _ = sw_class.get_or_create(
                 session,
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3976,13 +3976,16 @@ class GraphDatabase(SQLBase):
         with self._session_scope() as session, session.begin(subtransactions=True):
             if is_external:
                 sw_class = ExternalSoftwareEnvironment
+                python_version = None
             else:
                 sw_class = SoftwareEnvironment
+                python_version = image_name.rsplit("-", maxsplit=1)[1]   # pyXX
+                python_version = python_version[2:3] + "." + python_version[3:]   # take first digit and put . after it
 
             software_environment, _ = sw_class.get_or_create(
                 session,
                 environment_name=environment_name,
-                python_version=None,  # TODO: find Python version which would be used by default
+                python_version=python_version,
                 image_name=image_name,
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -3974,13 +3974,16 @@ def sync_analysis_result(self, document: dict) -> None:
         with self._session_scope() as session, session.begin(subtransactions=True):
             if is_external:
                 sw_class = ExternalSoftwareEnvironment
+                python_version = None
             else:
                 sw_class = SoftwareEnvironment
+                python_version = image_name.rsplit("-", maxsplit=1)[1]   # pyXX
+                python_version = python_version[2:3] + "." + python_version[3:]   # take first digit and put . after it
```
fridex: LGTM.

Just please add some checks here so we fail with corresponding error message. Otherwise, we can sync wrong data.

KPostOffice: Do you want me to raise an error or just _LOGGER.error(...) and return?

fridex: raise an error so the transaction is rolled back and no data are persisted

KPostOffice:
@fridex PTAL

thoth/storages/graph/postgres.py
```diff
@@ -3974,13 +3975,18 @@ def sync_analysis_result(self, document: dict) -> None:
...
+                if not re.match(r"^\d\.\d*$", python_version):
```
fridex: I think the star should not be present:

```python
>>> import re
>>> re.match(r"^\d\.\d*$", "3.")
<re.Match object; span=(0, 2), match='3.'>
>>> re.match(r"^\d\.\d*$", "3.14159265359")
<re.Match object; span=(0, 13), match='3.14159265359'>
>>> re.match(r"^\d\.\d$", "3.14159265359")
>>> re.match(r"^\d\.\d$", "3.")
>>>
```

KPostOffice: I don't understand why not?  Isn't it possible to have a python version `3.14149`?  The other option is to have something more like this:
```python
>>> import re
>>> re.match(r"^\d(\.\d+)*$", "3.")
>>> re.match(r"^\d(\.\d+)*$", "3")
<re.Match object; span=(0, 1), match='3'>
```

fridex: Agh, I missed this one. The issue is with "3." - we require a minor version to be present.

PR #1590, opened 2020-02-11
Added option to exclude models from the generated schema

CermakM:
Signed-off-by: Marek Cermak <macermak@redhat.com>

modified:   thoth/storages/cli.py

…

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

option to exclude models from the generated schema

### DIFF
--- a/thoth/storages/cli.py
+++ b/thoth/storages/cli.py
@@ -98,7 +98,7 @@ def generate_schema(schema_file: str, include=None, include_performance_models=T
     if include:
         desc_models = [
             m for m in all_models
-            if any(re.search(p, m.__name__) for p in include)
+            if any(re.fullmatch(p, m.__name__) for p in include)
         ]
     else:
         desc_models = all_models
--- a/thoth/storages/cli.py
+++ b/thoth/storages/cli.py
@@ -18,6 +18,7 @@
 """A CLI client to thoth-storages library."""
 
 import logging
+import re
 
 import click
 import daiquiri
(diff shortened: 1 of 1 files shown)


thoth/storages/cli.py
```diff
@@ -61,21 +62,52 @@ def cli(ctx=None, verbose: bool = False):
...
 
+    from thoth.storages.graph.models import ALL_MAIN_MODELS
+    from thoth.storages.graph.models import ALL_RELATION_MODELS
+
+    all_models = list(ALL_MAIN_MODELS.union(ALL_RELATION_MODELS))
+
+    if include:
+        desc_models = [
+            m for m in all_models
+            if any(re.search(p, m.__name__) for p in include)
```
fridex: I would probably expect `re.fullmatch` here. WDYT?

CermakM: You're probably right. Although it is more convenient to have just the `re.search` here (consider the following example), with `re.fullmatch` we avoid unexpected results.
```json
"args": [
                "generate-schema",
                "--include",
                "'^Dependency'",
                "--include",
                "'^Inspection'",
                "--include",
                "'^Hardware'",
                "--include",
                "'^Software'",
                "--include",
                "'SoftwareStack'",
                "inspection-schema-update.png"
]
```
would have to become
```json
            "args": [
                "generate-schema",
                "--include",
                "'^Dependency(.*)'",
                "--include",
                "'^Inspection(.*)'",
                "--include",
                "'^Hardware(.*)'",
                "--include",
                "'^Software(.*)'",
                "--include",
                "'(.*)SoftwareStack(.*)'",
                "inspection-schema-update.png"
            ]
```

PR #1592, opened 2020-02-12
Add check in sync when report is not provided by Adviser

pacospace:

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Add check in adviser sync when report is not provided by Adviser document

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -4548,6 +4548,12 @@ class GraphDatabase(SQLBase):
             )
 
             # Output stacks - advised stacks
+            if not document["result"].get("report", {}):
+                _LOGGER.warning(
+                    "No report found in %r", adviser_document_id
+                )
+                return
+
             for idx, product in enumerate(document["result"].get("report", {}).get("products", [])):
                 print(idx)
                 performance_score = None


thoth/storages/graph/postgres.py
```diff
@@ -4548,46 +4548,51 @@ def sync_adviser_result(self, document: dict) -> None:
...
-                performance_score = None
-                overall_score = product["score"]
-                for entry in product.get("justification", []):
-                    if "performance_score" in entry:
-                        if performance_score is not None:
-                            _LOGGER.error(
-                                "Multiple performance score entries found in %r (index: %d)", adviser_document_id, idx
-                            )
-                        performance_score = entry["performance_score"]
+            if document["result"].get("report", {}):
```
fridex: To reduce code complexity:

```python
            if not document["result"].get("report"):
                return
```

pacospace: Thanks! and leave the warning?

fridex: Yes. :+1:

thoth/storages/graph/postgres.py
```diff
@@ -4548,6 +4548,12 @@ def sync_adviser_result(self, document: dict) -> None:
             )
 
             # Output stacks - advised stacks
+            if not document["result"].get("report", {}):
```
fridex: Also, the sync of unsolved packages will need to go before this if statement - there is no report created if there are unsolved packages found by resolver.

pacospace: But it depends on the same `document["result"].get("report", {})`
```
            # Mark down packages that were not solved if adviser run failed.
            for unresolved in document["result"].get("report", {}).get("_ERROR_DETAILS", {}).get("unresolved", []):
                self._create_python_package_version(
                    session,
                    package_name=unresolved,
                    package_version=None,
                    index_url=None,
                    os_name=None,
                    os_version=None,
                    python_version=None,
                    sync_only_entity=True,
                )
```

pacospace: But it depends on the same `document["result"].get("report", {})`

```
            # Mark down packages that were not solved if adviser run failed.
            for unresolved in document["result"].get("report", {}).get("_ERROR_DETAILS", {}).get("unresolved", []):
                self._create_python_package_version(
                    session,
                    package_name=unresolved,
                    package_version=None,
                    index_url=None,
                    os_name=None,
                    os_version=None,
                    python_version=None,
                    sync_only_entity=True,
                )
```

Issue #1586, opened 2020-02-10
sqlalchemy.exc.InternalError: (psycopg2.errors.DataCorrupted)


KPostOffice:
**Describe the bug**

**To Reproduce**
call `graph.get_python_package_hashes_sha256("setuptools", "0.6b4", "https://pypi.org/simple")`
**Expected behavior**
Returns a list of package hashes for "setuptools-0.6b4"

**Screenshots**
```bash
DEBUG:https://pypi.org:443  "GET /pypi/setuptools/0.6b4/json HTTP/1.1" 200 139978
Traceback (most recent call last):
  File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1246, in _execute_context
    cursor, statement, parameters, context
  File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 588, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.DataCorrupted: could not read block 412 in file "base/12406/41007": read only 0 of 8192 bytes
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "app.py", line 80, in <module>
    main()
  File "app.py", line 71, in main
    stored_hashes = sorted(graph.get_python_package_hashes_sha256(pkg_ver[0], pkg_ver[1], pkg_ver[2]))
  File "/opt/app-root/lib/python3.6/site-packages/thoth/storages/graph/postgres.py", line 2635, in get_python_package_hashes_sha256
    result = query.all()
  File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/orm/query.py", line 3233, in all
    return list(self)
  File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/orm/query.py", line 3389, in __iter__
    return self._execute_and_instances(context)
  File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/orm/query.py", line 3414, in _execute_and_instances
    result = conn.execute(querycontext.statement, self._params)
  File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 982, in execute
    return meth(self, multiparams, params)
  File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/sql/elements.py", line 293, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
[...]

PR #1593, opened 2020-02-12
All counts optional

KPostOffice:
#1586 

Allow count to be optional to remove limit.

`.limit(None)` evaluates to no limit and will return all.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -403,7 +403,7 @@ class GraphDatabase(SQLBase):
             }
 
     def _do_software_environment_listing(
-        self, start_offset: int, count: int, is_external: bool, environment_type: str
+        self, start_offset: int, count: Optional[int], is_external: bool, environment_type: str
     ) -> List[str]:
         """Perform actual query to software environments."""
         if is_external:
@@ -423,7 +423,7 @@ class GraphDatabase(SQLBase):
             return [item[0] for item in result]
 
     def get_run_software_environment_all(
-        self, start_offset: int = 0, count: int = DEFAULT_COUNT, is_external: bool = False
+        self, start_offset: int = 0, count: Optional[int] = DEFAULT_COUNT, is_external: bool = False
     ) -> List[str]:
         """Get all software environments available for run.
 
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -453,7 +455,7 @@ def _do_software_environment_analyses_listing(
         self,
         software_environment_name: str,
         start_offset: int,
-        count: int,
+        count: Optional[int],
```
fridex: Could you point me to docs where `limit` can be set to `None`? I tried to look up this behaviour, but failed:

[1] https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.limit
[2] https://github.com/sqlalchemy/sqlalchemy/blob/a54527fb0ef71aca0be9614617f143da27e03f22/lib/sqlalchemy/sql/selectable.py#L2669-L2670

KPostOffice: [3] https://stackoverflow.com/questions/12836194/how-do-i-remove-a-limit-that-is-already-applied-to-a-query-object

I found this

KPostOffice: And also, default value of limit is `None`
[4] https://github.com/sqlalchemy/sqlalchemy/blob/a54527fb0ef71aca0be9614617f143da27e03f22/lib/sqlalchemy/sql/selectable.py#L2457-L2462

fridex: I'm not that happy about merging officially undocumented functionality as it might break with updates. As this is sqlalchemy, let's keep eyes closed and assume their api is stable.

fridex: > And also, default value of limit is `None`
> [4] https://github.com/sqlalchemy/sqlalchemy/blob/a54527fb0ef71aca0be9614617f143da27e03f22/lib/sqlalchemy/sql/selectable.py#L2457-L2462

But checked before applying it:

https://github.com/sqlalchemy/sqlalchemy/blob/a54527fb0ef71aca0be9614617f143da27e03f22/lib/sqlalchemy/sql/selectable.py#L2482-L2483

PR #1604, opened 2020-02-17
Parse py version

KPostOffice:

see https://github.com/thoth-station/storages/pull/1577#discussion_r380245363

This will match minor and major releases and not strings with errant periods at the end.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3980,7 +3980,7 @@ class GraphDatabase(SQLBase):
                 sw_class = SoftwareEnvironment
                 python_version = image_name.rsplit("-", maxsplit=1)[1]   # pyXX
                 python_version = python_version[2:3] + "." + python_version[3:]   # take first digit and put . after it
-                if not re.match(r"^\d\.\d*$", python_version):
+                if not re.match(r"^\d(\.\d+)*$", python_version):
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -3982,7 +3982,7 @@ def sync_analysis_result(self, document: dict) -> None:
                 sw_class = SoftwareEnvironment
                 python_version = image_name.rsplit("-", maxsplit=1)[1]   # pyXX
                 python_version = python_version[2:3] + "." + python_version[3:]   # take first digit and put . after it
-                if not re.match(r"^\d\.\d*$", python_version):
+                if not re.match(r"^\d(\.\d+)*$", python_version):
```
fridex: We expect both to be present: `r"^\d\.\d+$"`

fridex: Actually, this will work because the dot will be always present (based on the code above) forcing the minor version to be checked. IMHO worth to be explicit here

KPostOffice: I'm unsure what you mean.  Should we make sure there is a minor version attached or not?

fridex: > I'm unsure what you mean. Should we make sure there is a minor version attached or not?

Yes, the minor version should always present. The major version as well.

PR #1611, opened 2020-02-19
flag for missing package version

KPostOffice:
Add flag which indicates whether a package version has gone missing from a python index.


thoth/storages/graph/models.py
```diff
@@ -68,6 +68,7 @@ class PythonPackageVersion(Base, BaseExtension):
     python_package_metadata_id = Column(
         Integer, ForeignKey("python_package_metadata.id", ondelete="CASCADE"), nullable=True
     )
+    is_missing = Column(Boolean, nullable=False, default=False)
```
pacospace: Please provide also the corresponding alembic migration :) https://github.com/thoth-station/storages/blob/master/README.rst#generating-migrations-and-schema-adjustment-in-deployment

PR #1617, opened 2020-02-24
Move import to local use

fridex:

With argo workflows, we don't have this function anymore.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


fridex:
I'll merge this as this blocks us from running dependency monkey. Feel free to raise any comments.

PR #1618, opened 2020-02-24
API calls for package-update-consumer

KPostOffice:

…

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

A few api calls for updating and querying the database for package-update-job


thoth/storages/graph/postgres.py
```diff
@@ -3771,6 +3771,41 @@ def create_python_cve_record(
...
+    def get_repositories_using_package(index_url: str, package_name:str):
+        with self._session_scope() as session, session.begin(subtransactions=True):
+            result = (
+                session.query(PythonPackageVersion)
+                .filter(index_url==index_url)
+                .filter(package_name==package_name)
+                .join(PythonRequirementsLock)
+                .join(ExternalSoftwareEnvironment)
+                .join(AdviserRun)
+                .with_entities(origin)
```
fridex: `distinct`?

For this one and `get_repositories_using_package_version` as well.

thoth/storages/graph/postgres.py
```diff
@@ -3771,6 +3771,41 @@ def create_python_cve_record(
...
+            )
+
+    def get_repositories_using_package_version():
```
fridex: Missing docstring, here and for `get_repositories_using_package`.

thoth/storages/graph/postgres.py
KPostOffice: Yeah that's probably best because adviser can be run multiple times under the same origin.

KPostOffice: Hmmmm, this will get old stacks as well...

KPostOffice: I made some changes... I think it should be closer to the intended functionality now.

thoth/storages/graph/postgres.py
```diff
@@ -3771,6 +3771,52 @@ def create_python_cve_record(
...
+        """Update value of is_missing flag for PythonPackageVersion."""
+        with self._session_scope():
+            (session.query(PythonPackageVersion)
+                .update()
+                .where(index_url == index_url)
+                .where(package_name == package_name)
+                .where(package_version == package_version)
+                .values(is_missing=value))
+
+    def get_repositories_using_package(index_url: str, package_name: str) -> List[str]:
```
pacospace: According to naming conventions for queries, this should have `_all` at the end. Moreover, in the docstring I would add `unique` urls of repos... becuase the `distinct` filter is not optional in this case. Same for `get_repositories_using_package_version`.

pacospace:
> LGTM, @pacospace PTAL.
> 
> BTW it would be also good to publish the naming convention here :)

There is a card on that in Next :D

harshad16:
recheck

harshad16:
recheck

thoth/storages/graph/postgres.py
```diff
@@ -3771,6 +3771,70 @@ def create_python_cve_record(
 
                 return False
 
+    def update_missing_flag_package_version(
+        self,
+        value: bool,
+        index_url: str,
+        package_name: str,
+        package_version: str
```
fridex: Positional arguments in this module are in the following order:

1. package name
2. package version
3. index url

Please move the `value` parameter at the end so it reads as "what needs to be updated with what".

thoth/storages/graph/postgres.py
```diff
@@ -3771,6 +3771,70 @@ def create_python_cve_record(
...
+
+    def get_all_repositories_using_package(self, index_url: str, package_name: str) -> List[str]:
+        """Given package info, get unique urls of repos which have been advised with this package."""
+        with self._session_scope() as session:
+            result = (
+                session.query(PythonPackageVersion)
+                .join(PythonPackageIndex)
+                .join(PythonRequirementsLock)
+                .join(AdviserRun, AdviserRun.user_software_stack_id == PythonRequirementsLock.python_software_stack_id)
+                .order_by(AdviserRun.origin)
```
fridex: Is this ordering required? Here and bellow.

KPostOffice: This one might not be, but the other one definitely is.  `.distinct` gets the first result so we want to make sure that each origin has only one result.  I am trying to ensure that the entry which we get corresponds to the most up to date information which we have on said `origin`.

PR #1635, opened 2020-03-02
Remove constraint on deployment name to allow for retrieval of inspections

pacospace:

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Remove the constraint on THOTH_DEPLOYMENT_NAME env variable. We were storing documents in 
```
self.prefix = "{}/{}/{}".format(
                prefix or os.environ["THOTH_CEPH_BUCKET_PREFIX"], self.deployment_name, self.RESULT_TYPE
            )
```
e.g. `thoth/data/thoth-psi-stage/inspections`
with `prefix=thoth/data`, `deployment_name=thoth-psi-stage`

but with argo workflows we store results in:
```
            self.prefix = "{}/{}".format(
                prefix or os.environ["THOTH_CEPH_BUCKET_PREFIX"], self.RESULT_TYPE
            )
```

e.g. `argo/inspections`
## Description

Relax constraint on environment variable and add an if statement to maintain the previous compatibility for bucket on Ceph

### DIFF
--- a/thoth/storages/result_base.py
+++ b/thoth/storages/result_base.py
@@ -55,10 +55,16 @@ class ResultStorageBase(StorageBase):
             self.RESULT_TYPE
         ), "Make sure RESULT_TYPE in derived classes to distinguish between adapter type instances is non-empty."
 
-        self.deployment_name = deployment_name or os.environ["THOTH_DEPLOYMENT_NAME"]
-        self.prefix = "{}/{}/{}".format(
-            prefix or os.environ["THOTH_CEPH_BUCKET_PREFIX"], self.deployment_name, self.RESULT_TYPE
-        )
+        self.deployment_name = deployment_name or os.getenv("THOTH_DEPLOYMENT_NAME")
+
+        if self.deployment_name:
+            self.prefix = "{}/{}/{}".format(
+                prefix or os.environ["THOTH_CEPH_BUCKET_PREFIX"], self.deployment_name, self.RESULT_TYPE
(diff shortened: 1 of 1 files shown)


pacospace:
recheck

fridex:
IMHO we should tweak argo inspections to place results to the correct bucket.

pacospace:
> IMHO we should tweak argo inspections to place results to the correct bucket.

That is what I thougth also, but we cannot analyze inspection results in the notebook at the moment then :)

fridex:
> > IMHO we should tweak argo inspections to place results to the correct bucket.
> 
> That is what I thougth also, but we cannot analyze inspection results in the notebook at the moment then :)

We can place it to the correct location. I turned off the scheduling cronjob and we have finished ubi-8/py36 inspections. We can start fedora ones in the next scheduling round - meanwhile these can be placed to the correct location - the old ones can be moved to some backup not to pollute buckets.

pacospace:
> > > IMHO we should tweak argo inspections to place results to the correct bucket.
> > 
> > 
> > That is what I thougth also, but we cannot analyze inspection results in the notebook at the moment then :)
> 
> We can place it to the correct location. I turned off the scheduling cronjob and we have finished ubi-8/py36 inspections. We can start fedora ones in the next scheduling round - meanwhile these can be placed to the correct location - the old ones can be moved to some backup not to pollute buckets.

Perfect then! I can close this PR. Thanks!

PR #1665, opened 2020-03-20
Link Adviser Run with Python Package Versio Entity

pacospace:

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Add New table HasUnresolved, which connect AdviserRun with PythonPackageVersionEntity. This is import to have queries that can retrieve unsolved packages for a certain adviser run, so that we can analyze unsolved packages and rerun adviser run in order to go back to the user.


thoth/storages/graph/models.py
```diff
@@ -545,6 +545,8 @@ class AdviserRun(Base, BaseExtension):
     debug = Column(Boolean, nullable=False)
     limit_latest_versions = Column(Integer, nullable=True)
     adviser_error = Column(Boolean, nullable=False, default=False)
+    adviser_error_details = Column(Text, nullable=True)
```
fridex: Do we want to keep these in this database?

pacospace: I was thinking that probably is not required in this case :) I'll remove it. Thanks!!

pacospace:
I realized we need a flag to know which adviser run we already rescheduled, otherwise, we keep rescheduling adviser runs. But what happens if adviser fails after?

Issue #1436, opened 2020-01-06
Use Text instead of String all over the place


fridex:
**Is your feature request related to a problem? Please describe.**

As there is no strict difference between Text and String datatypes in postgres, let's maintain just Text types.

PR #1668, opened 2020-03-23
Modify schema and logic to sync adviser run

pacospace:

Fixes: https://github.com/thoth-station/storages/issues/1436

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

## This Pull Request 

- Substitute String with Text in all models.
- adds two parameters required to deal with unsolved packages in adviser run. 
- modifies syncing logic of adviser document to cope with the cases on unsolved dependencies.
- adds query that will be used by the consumer that will re run the adviser run for the users.


thoth/storages/graph/models.py
```diff
@@ -543,6 +543,8 @@ class AdviserRun(Base, BaseExtension):
     origin = Column(String(256), nullable=True)
     is_s2i = Column(Boolean, nullable=True)
     debug = Column(Boolean, nullable=False)
+    need_re_run = Column(Boolean, nullable=True)
+    re_run_adviser_id = Column(String(256), nullable=True)
```
fridex: I think we can stick with Text, PostgreSQL treats these two with the same underlying logic so no downside when using Text.

thoth/storages/graph/postgres.py
```diff
@@ -3331,6 +3331,42 @@ def get_python_package_version_metadata(
...
+            )
+
+            query_result = query.all()
+
+            result = {}
+            for couple in query_result:
+                if couple[0] not in result.keys():
+                    result[couple[0]] = []
+
+                result[couple[0]].append(couple[1])
```
fridex: What about something like this one?

```python
result.setdefault(couple[0], []).append(couple[1])
```

thoth/storages/graph/postgres.py
```diff
@@ -3331,6 +3331,42 @@ def get_python_package_version_metadata(
...
+
+            if result is None:
+                logger.info(f"No Adviser Run found that need to be re run!")
```
fridex: This message could be more suitable for consumer, I would rather keep this log message on the component level.

thoth/storages/graph/postgres.py
```diff
@@ -4535,37 +4572,184 @@ def sync_adviser_result(self, document: dict) -> None:
...
-            )
-
             # Output stacks - advised stacks
             if not document["result"].get("report", {}):
                 _LOGGER.warning(
                     "No report found in %r", adviser_document_id
                 )
+                adviser_run, _ = AdviserRun.get_or_create(
+                    session,
+                    additional_stack_info=bool(document["result"].get("stack_info")),
```
fridex: To reduce copying this 4 times, you can create a dict and use kwargs to propagate values:

```python
attributes = {
  "additional_stack_info": bool(document["result"].get("stack_info")),
  "advised_configuration_changes": bool(document["result"].get("advised_configuration")),
  # and so on...
}
```

```
AdviserRun.get_or_create(**properties)
```

Or, if you want to distinguish cases where it changes:

```
 AdviserRun.get_or_create(**attributes, need_re_run=False)  # need_re_run must not be in attributes
```

Another solution would be to add items to `attributes` conditionally:

```python
if unresolved_packages:
    attributes["need_re_run"] = True
```
    


IMHO it will reduce code size and could be less error prone.

thoth/storages/graph/models.py
pacospace: Right we have to move all to Text actually :) Thank!

thoth/storages/graph/postgres.py
```diff
@@ -4535,37 +4566,116 @@ def sync_adviser_result(self, document: dict) -> None:
...
+                    session.query(AdviserRun)
+                    .filter(AdviserRun.adviser_document_id == re_run_adviser_id)
+                    .first()
+                )
+                # INSERT…ON CONFLICT (Upsert)
+                # https://docs.sqlalchemy.org/en/13/dialects/postgresql.html?highlight=conflict#insert-on-conflict-upsert
+                # https://docs.sqlalchemy.org/en/13/errors.html#sql-expression-language compile required
+                if first_adviser_run and first_adviser_run.need_re_run:
+                    insert_stmt = insert(AdviserRun).values(
+                        id=first_adviser_run.id,
```
fridex: Something like this to avoid enumerating all the properties (when they change):

```python
insert_stmt = insert(AdviserRun).values(**first_adviser_run.to_dict(without_id=False))
```

pacospace: Much better :) Thanks!

PR #1669, opened 2020-03-23
Revert "API calls for package-update-consumer"

KPostOffice:
Reverts thoth-station/storages#1618


KPostOffice:
Accidentally merged

thoth/storages/graph/postgres.py
```diff
@@ -3775,97 +3775,6 @@ def create_python_cve_record(
...
-                .join(HasArtifact)
-                .join(PythonArtifact)
-                .filter(PythonArtifact.artifact_hash_sha256 == sha256_hash)
-                .with_entities(PythonArtifact.id)
-            )
-            # Can a hash be present on more than one python_version_entity?
-            (
-                session.query(PythonArtifact)
-                .filter(PythonArtifact.id.in_(subq))
-                .delete(synchronize_session='fetch')
```
fridex: We should be careful with deletions. We have 'on delete cascade' in our database that, in the worst case, can cause deleting the whole database when deleting a single row. Might be good to consider "active" flag. In that case we would be able to tell people "yes, we saw this hash before, but the artifact with that hash was deleted" that can be valuable info.

PR #1671, opened 2020-03-24
Package update api

KPostOffice:

…

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

See #1618


thoth/storages/graph/postgres.py
```diff
@@ -3775,6 +3775,98 @@ def create_python_cve_record(
...
+                    .filter(PythonPackageVersion.package_name == package_name)
+                    .filter(PythonPackageVersion.package_version == package_version)
+                    .filter(PythonPackageIndex.url == index_url)
+                    .with_entities(PythonPackageVersion.id)
+                    )
+            (
+                session.query(PythonPackageVersion)
+                .filter(PythonPackageVersion.id.in_(subq))
+                .update({"is_missing": value}, synchronize_session='fetch')
+            )
```
pacospace: Don't we want `upsert` functionality in this case?
Have a look at https://github.com/thoth-station/storages/blob/158d7a368cc25e984c23080ff02f91cbc7a7827f/thoth/storages/graph/postgres.py#L4646

thoth/storages/graph/postgres.py
```diff
@@ -3775,6 +3775,98 @@ def create_python_cve_record(
...
+
+    def get_all_repositories_using_package(self, package_name: str, index_url: str) -> List[str]:
```
pacospace: - `origin` in general encompass more than just repositories.
- `package` is not accepted, in general we refer to `python_package` for `PythonPackageVersionEntity` and `python_package_version` for `PythonPackageVersion`

Please follow query name conventions: https://github.com/thoth-station/storages/tree/master/docs/conventions and if you think a new case should be added please don't hesitate to open a PR :) 

The closest one I see in this case should be `get_adviser_run_origin_all_python_package_version`, `get_<objects>_all_<attribute_filter>()`, but maybe there is a new for a new case as this queries have several joins.

Also in the query convention there is missing the need to add to the doc string of each query an example. See below:
```
        """Retrieve all origins (e.g. repos URLs) in adviser run for a specific Python Package Version.

        Examples:
        >>> from thoth.storages import GraphDatabase
        >>> graph = GraphDatabase()
        >>> graph.get_adviser_run_origin_all_python_package_version(package_name='numpy', index_url='https://pypi.org/simple')
        [HERE SHOULD GO AN EXAMPLE OF A RESULT]
        """
```

thoth/storages/graph/postgres.py
```diff
@@ -3775,6 +3775,98 @@ def create_python_cve_record(
...
+                .order_by(AdviserRun.origin)
+                .order_by(AdviserRun.datetime.desc())
+                .distinct(AdviserRun.origin)
+                .filter(PythonPackageIndex.url == index_url)
+                .filter(PythonPackageVersion.package_name == package_name)
+                .filter(PythonPackageVersion.package_version == package_version)
+                .with_entities(AdviserRun.origin)
+            ).all()
+            result = [x[0] for x in result]
+            return result
```
pacospace: This query is redundant and can be merged with the one above, adding an optional filter for the version if present :)

thoth/storages/graph/postgres.py
KPostOffice: This should never insert into the table. Not sure if that matters.

thoth/storages/graph/postgres.py
KPostOffice: What else does origin encompass? I thought we were only using it when we trigger advises to indicate what repository it was triggered by.

thoth/storages/graph/postgres.py
pacospace: You want to update an existing record in the database, by default that flag is Null right now I guess, but if you want to insert a value on existing record without creating a new record, you need to use upsert, or insert on conflict. Update only will create a new record as far as I'm aware of and from I understood we don't want a new record but changing a value of an existing one.

thoth/storages/graph/postgres.py
```diff
@@ -3775,6 +3775,82 @@ def create_python_cve_record(
...
+                session.query(PythonPackageVersion)
+                .join(PythonPackageIndex)
+                .join(PythonRequirementsLock)
+                .join(AdviserRun, AdviserRun.user_software_stack_id == PythonRequirementsLock.python_software_stack_id)
+                .order_by(AdviserRun.origin)
+                .order_by(AdviserRun.datetime.desc())
+                .distinct(AdviserRun.origin)
+                .filter(PythonPackageIndex.url == index_url)
+                .filter(PythonPackageVersion.package_name == package_name)
+                .filter(PythonPackageVersion.package_version == package_version)
```
pacospace: This should not be here if you put optional filter below.

thoth/storages/graph/postgres.py
pacospace: Actually I got confused, it's only git repos :) my bad. In any case to be coherent with conventions and queries in general: `get_adviser_run_origins_all` should retrieve all origins in adviser run, then filter can be applied: package_name, package_version, index_url, distinct also should be there because if we want to count how many for a specific origin we can reuse logic across queries.

thoth/storages/graph/postgres.py
```diff
@@ -3775,6 +3775,82 @@ def create_python_cve_record(
...
+            )
+            if package_version is not None:
+                result.filter(PythonPackageVersion.package_version == package_version)
+
+            result = result.with_entities(AdviserRun.origin).all()
+            result = [x[0] for x in result]
+            return result
```
pacospace: In any case to be coherent with conventions and queries in general: - `get_adviser_run_origins_all` should retrieve all origins in adviser run, then filters can be applied: 

- `package_name`, `package_version`, `index_url`, `distinct` also should be there because if we want to count how many for a specific origin we can reuse logic across queries. 

- `start_offset` and `count` should be considered if we want to consider pagination purposes.

Moreover `package_name` and `package_version` should be used with normalize method:

`package_name = self.normalize_python_package_name(package_name)`
`package_version = self.normalize_python_package_version(package_version)`

What about?

```
    def get_adviser_run_origins_all(
        self,
        *,
        package_name: str = None,
        package_version: str = None,
        index_url: str = None,
        start_offset: int = 0,
        count: Optional[int] = DEFAULT_COUNT,
        distinct: bool = False,
    ) -> List[str]:
        """Retrieve all origins (git repos URLs) in Adviser Run.

        Examples:
        >>> from thoth.storages import GraphDatabase
        >>> graph = GraphDatabase()
        >>> graph.get_adviser_run_origins_all()
        [HERE SHOULD GO AN EXAMPLE OF A RESULT]
        """

            query = (
                session.query(AdviserRun)
                .order_by(AdviserRun.origin)
                .order_by(AdviserRun.datetime.desc())
            )

            if package_version or package_name:
                query = (
                    query.join(PythonRequirementsLock, PythonRequirementsLock.python_software_stack_id == AdviserRun.user_software_stack_id)
                    .join(PythonPackageVersion)
                )

                if index_url:
                    query = query.join(PythonPackageIndex)

            query = query.with_entities(AdviserRun.origin)

            if package_name is not None:
                query.filter(PythonPackageVersion.package_name == package_name)

            if package_version is not None:
                query.filter(PythonPackageVersion.package_version == package_version)

            if index_url is not None:
                query.filter(PythonPackageIndex.url  == index_url)

            query = query.offset(start_offset).limit(count)

            if distinct:
                query = query.distinct()

            result = [r[0] for r in query.all() if r[0]] # We do not consider None results

            return result
```

pacospace: @fridex PTAL :)

thoth/storages/graph/postgres.py
KPostOffice: The default value should be false, I think we set it to not be nullable in the configuration

pacospace: > The default value should be false, I think we set it to not be nullable in the configuration

Then we need the upsert instead of the update, otherwise you will create a new record in the table and not updating the existing one.

thoth/storages/graph/postgres.py
```diff
@@ -3775,6 +3775,112 @@ def create_python_cve_record(
...
+
+            if package_name is not None:
+                package_name = self.normalize_python_package_name(package_name)
+                query.filter(PythonPackageVersion.package_name == package_name)
+
+            if package_version is not None:
+                package_version = self.normalize_python_package_version(package_version)
+                query.filter(PythonPackageVersion.package_version == package_version)
+
+            if index_url is not None:
```
pacospace: This should be also under `if package_name or package_version:` otherwise there would be a missing join.

thoth/storages/graph/postgres.py
```diff
@@ -3775,6 +3775,112 @@ def create_python_cve_record(
...
+
+            if package_name or package_version:
+                query = (
+                    query.join(
+                        ExternalPythonRequirementsLock,
+                        ExternalPythonRequirementsLock.python_software_stack_id == AdviserRun.user_sotware_stack_id
+                    )
+                    .join(PythonPackageVersion)
+                )
+                if index_url:
```
pacospace: `if index_url is not None:`

thoth/storages/data/alembic/versions/f6d46e0b37da_flag_for_if_specific_hash_has_gone_.py
```diff
@@ -0,0 +1,28 @@
...
+Create Date: 2020-03-24 21:32:57.426880+00:00
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'f6d46e0b37da'
+down_revision = '74254b596f00'
```
pacospace: Please update the revision references in the file, as another PR that changes the schema has been merged already: https://github.com/thoth-station/storages/pull/1668/files

therefore this change in schema should down revision: `0ea53ec16d6b` (https://github.com/pacospace/storages/blob/16d6929a76bdbf1961ac1d39c768bad445aef4a4/thoth/storages/data/alembic/versions/0ea53ec16d6b_use_text_instead_of_strings_all_over_.py#L3)

KPostOffice: I did look at upsert.  I still think update is what we want because we give the flag a default value so we should never be inserting.

thoth/storages/data/alembic/versions/ecb0ca258835_flag_for_if_specific_hash_has_gone_.py
```diff
@@ -0,0 +1,47 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('python_artifact', sa.Column('present', sa.Boolean(), nullable=False, server_default="false"))
+
+
+    # NOTE: I am unsure where these came from?
```
pacospace: This is happening because you are changing the datatype of `artifact_hash_sha256` and `artifact_name` from Text to String, why that? We want to have Text and do not use String actually.

PR #1691, opened 2020-03-28
Drop methods tools to gain performance

fridex:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


fridex:
```
thoth.adviser.resolver:1091: Resolver took 172.263 seconds in total, pipeline speed 290.254 stacks/second
```

versus

```
thoth.adviser.resolver:1091: Resolver took 113.202 seconds in total, pipeline speed 441.69 stacks/second
```

on a tested stack, scoring 50k stacks (the perf impact grows as the cache grows).

fridex:
The bottleneck seems to be wirerope and their mixins.

PR #1725, opened 2020-04-21
Introduce get_dependents query

fridex:

Needed for Thoth's reverse solver.

- [ ] Yes
- [x] No


thoth/storages/graph/postgres.py
```diff
@@ -2425,6 +2425,76 @@ def get_python_environment_marker_evaluation_result(
 
             return result[0]
 
+    def get_dependents(
```
pacospace, line 2428: `get_python_package_version_dependents_all`?

thoth/storages/graph/postgres.py
```diff
@@ -2425,6 +2425,76 @@ def get_python_environment_marker_evaluation_result(
...
+        os_version = self.normalize_os_version(os_name, os_version)
+
+        with self._session_scope() as session:
+            query = (
+                session.query(PythonPackageVersionEntity)
+                .filter(PythonPackageVersionEntity.package_name == package_name)
+                .join(DependsOn)
+            )
+
+            query = query.join(PythonPackageVersion)
```
pacospace, line 2463: With current implementation seems to be quite slow. Moreover I don't see results:

```
Test

FUNCTION: get_dependents

INPUTS: {'count': 100, 'start_offset': 0, 'python_version': None, 'os_version': None, 'os_name': None, 'package_name': 'thoth-worker'}

RESULTS: []
[]
11.01444746996276
```

Also isn't PPV enough? I tested with the following change:

```
            query = (
                session.query(PythonPackageVersion)
                .filter(PythonPackageVersion.package_name == package_name)
                .join(DependsOn)
            )

```

And obtained:
```
Test

FUNCTION: get_dependents

INPUTS: {'count': 100, 'start_offset': 0, 'python_version': None, 'os_version': None, 'os_name': None, 'package_name': 'thoth-worker'}

RESULTS: [{'package_name': 'thoth-worker', 'package_version': '0.0.2', 'index_url': 'https://pypi.org/simple', 'version_range': '*', 'marker_evaluation_result': True, 'marker': None, 'extra': None}]
[{'package_name': 'thoth-worker', 'package_version': '0.0.2', 'index_url': 'https://pypi.org/simple', 'version_range': '*', 'marker_evaluation_result': True, 'marker': None, 'extra': None}]
0.015326200984418392
```

It seems x10 faster

thoth/storages/graph/postgres.py
```diff
@@ -2425,6 +2425,76 @@ def get_python_environment_marker_evaluation_result(
...
+        >>> graph = GraphDatabase()
+        >>> graph.get_dependents("selinon", os_name="rhel", os_version="8", python_version="3.6")
+        [
+          {
+            "index_url": "https://pypi.org/simple",
+            "package_name": "thoth-worker",
+            "package_version": "0.0.2",
+            "version_range": ">=1.0.0",
+          }
+        ]
```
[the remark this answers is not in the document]
fridex: Good point, I forgot to reflect recent changes. :+1:

thoth/storages/graph/postgres.py
fridex: > Also isn't PPV enough?

Will it produce desired results? The query is reversed `get_depends_on` query where we get dependencies of a package. This query receives dependents, so I think we need to start in the entity and check what PPVs link to it.

> With current implementation seems to be quite slow. 

Sadly yes. After profiling it a bit, the query spends most of the time on `DISTINCT` where it is sorting and removing duplicates in 0.6M entries (using a stage dump). The good thing is that the reverse solver running this in the cluster will not consume much resources (in comparison to the solver).

> Moreover I don't see results.

How did you invoke the method? What parameters?

pacospace: > > Also isn't PPV enough?
> 
> Will it produce desired results? The query is reversed `get_depends_on` query where we get dependencies of a package. This query receives dependents, so I think we need to start in the entity and check what PPVs link to it.

Thanks :) 
> 
> > With current implementation seems to be quite slow.
> 
> Sadly yes. After profiling it a bit, the query spends most of the time on `DISTINCT` where it is sorting and removing duplicates in 0.6M entries (using a stage dump). The good thing is that the reverse solver running this in the cluster will not consume much resources (in comparison to the solver).

very nice 
> 
> > Moreover I don't see results.
> 
> How did you invoke the method? What parameters?

I retested with same parameters as in the example provided in the query method.

```
Test

FUNCTION: get_dependents

INPUTS: {'count': 100, 'start_offset': 0, 'python_version': '3.6', 'os_version': '8', 'os_name': 'rhel', 'package_name': 'selinon'}

RESULTS: [{'package_name': 'thoth-worker', 'package_version': '0.0.2', 'index_url': 'https://pypi.org/simple', 'version_range': '*', 'marker_evaluation_result': True, 'marker': None, 'extra': None}]
11.16158254200127
```

It works, I was using package with no dependents, but that also highlight the issue to have desired results

thoth/storages/graph/postgres.py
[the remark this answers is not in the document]
fridex: Done in 0ac8bbd

thoth/storages/graph/postgres.py
pacospace: what about the name @fridex ?

PR #1759, opened 2020-05-13
Add query to count a table

pacospace:

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Add query to count a specific table.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -5118,7 +5118,7 @@ class GraphDatabase(SQLBase):
         return result
 
     def get_entity_count(self, entity: Union[Base, BaseExtension]) -> int:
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -5115,6 +5117,13 @@ def get_pi_count(self, component: str) -> Dict[str, int]:
 
         return result
 
+    def get_entity_count(self, entity: Union[Base, BaseExtension]) -> int:
```
goern, line 5120: I think the docstring is not good, what does this method do?

pacospace: Updated

pacospace:
recheck

PR #1765, opened 2020-05-14
function for checking current availability of package

KPostOffice:
This checks whether a specific python package version has gone missing from an index.  This is required for implementing a sieve to not include missing packages in adviser

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3933,6 +3933,22 @@ class GraphDatabase(SQLBase):
                 .update({"is_missing": value}, synchronize_session='fetch')
             )
 
+    def is_python_package_version_missing(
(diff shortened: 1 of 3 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -3933,6 +3933,25 @@ def update_missing_flag_package_version(
                 .update({"is_missing": value}, synchronize_session='fetch')
             )
 
+    def is_python_package_version_is_missing(
+        self,
+        package_name: str,
+        package_version: str,
+        index_url: str,
+    ) -> bool:
+        """Check whether is missing flag is set for python package version."""
```
pacospace, line 3942: `is_missing`

harshad16:
recheck

PR #1770, opened 2020-05-18
Add is_missing optional argument to all pypackageversion queries

KPostOffice:
Add optional is_missing param when querying for python_package_versions.  We may want to make this default behavior instead of defaulting to `None`.  Let me know what you think.


KPostOffice:
recheck

goern:
recheck

harshad16:
recheck

PR #1803, opened 2020-06-08
create class for storing security indicators

KPostOffice:
Store security indicator results


thoth/storages/graph/security_indicators.py
```diff
@@ -0,0 +1,26 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Adapter for storing security indicators (SI) onto a persistence remote store."""
+
+from .result_base import ResultStorageBase
+
+
+class SecurityIndicatorResultsStore(ResultStorageBase):
+    """Adapter for persisting recommendation results."""
```
pacospace, line 24: `"""Adapter for persisting security indicators results."""

pacospace:
> Do we want to have specific adapters per security indicator result type? (bandit, cloc, ..)

I thought everything could go to one security-indicator, but they produce different outputs probably difficult to be generalized. So maybe it's a good idea to have separate adapters.

thoth/storages/si_bandit.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Adapter for storing bandit security indicators (SI) onto a persistence remote store."""
+
+from .result_base import ResultStorageBase
+
+
+class SecurityIndicatorResultsStore(ResultStorageBase):
```
pacospace, line 23: class names should be changed as well :)

fridex: And please [add them also to the `__init__.py` file](https://github.com/thoth-station/storages/blob/master/thoth/storages/__init__.py) so we can easily import them.

fridex:
recheck

PR #1836, opened 2020-06-18
Introduce a method for checking if the given inspection exists

fridex:
# This introduces a breaking change

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


fridex:
regate

PR #1859, opened 2020-06-26
WIP: Add source

saisankargochhayat:
## Related Issues and Dependencies
Closes #1850

PR #1862, opened 2020-06-29
New enum

saisankargochhayat:
Related to #1859 
Added new enum and auto-generated alembic upgrade still includes all column


thoth/storages/graph/enums.py
```diff
@@ -70,3 +70,13 @@ class QuerySortTypeEnum(Enum):
 
     PACKAGE_NAME = "package_name"
     PACKAGE_VERSION = "package_version"
+
+
+class ThothAdviserIntegrationEnum(Enum):
+    """Class for source type enum."""
+
+    CLI = "CLI"
```
pacospace, line 78:
```
from enum import auto
```
you can use `auto()`.
example `CLI = auto()`

saisankargochhayat: `ENUM` class doesn't support auto(), that is if you pass a enum class with as above, it seems to give an error like this - https://github.com/thoth-station/storages/pull/1859#discussion_r447112712

thoth/storages/data/alembic/versions/76413cbe99bf_added_source_type_to_adviser_run_table.py
```diff
@@ -0,0 +1,72 @@
...
+revision = '76413cbe99bf'
+down_revision = '7c261487d4d4'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('adviser_run', sa.Column('source_type', postgresql.ENUM('CLI', 'KEBECHET', 'S2I', 'GITHUB_APP', 'JUPYTER_NOTEBOOK', name='source_type'), nullable=True))
+    op.create_foreign_key(None, 'adviser_run', 'external_software_environment', ['external_run_software_environment_id'], ['id'], ondelete='CASCADE')
```
saisankargochhayat, line 22: @fridex seems like its an alembic issue - https://github.com/sqlalchemy/alembic/issues/363#issuecomment-614962398
Should I just manually remove these?

fridex: > Should I just manually remove these?

Yes, please do so.

thoth/storages/graph/enums.py
fridex: Based on the error:

```
TypeError: object of type 'int' has no len()
```

you've probably passed value of the enum to the `source_type` enum on postgres side:

```python
    source_type = Column(
        ENUM(
            ThothAdviserIntegrationEnum.CLI.value,
            ThothAdviserIntegrationEnum.KEBECHET.value,
            ThothAdviserIntegrationEnum.S2I.value,
            ThothAdviserIntegrationEnum.GITHUB_APP.value,
            ThothAdviserIntegrationEnum.JUPYTER_NOTEBOOK.value,
            name="source_type",
            create_type=True,
        ),
        nullable=True,
    )
```

That will turn name into an `int` based on `auto()` invocation. You can provide enum names (using this trick should not enforce their enumeration every time):

```python
    source_type = Column(
        ENUM(
            *(e.name for e in ThothAdviserIntegrationEnum),
            name="source_type",
            create_type=True,
        ),
        nullable=True,
    )
```

saisankargochhayat: Thanks! auto() does seem to work now after passing `.name`. 
> (using this trick should not enforce their enumeration every time):
Sorry I didn't understand how this would affect the enumeration?

thoth/storages/data/alembic/versions/76413cbe99bf_added_source_type_to_adviser_run_table.py
saisankargochhayat: I tried removing the unnecessary migration statments, still seems to failing..
Any suggestion on this - https://github.com/sqlalchemy/alembic/issues/363#issuecomment-614962398
Should we have just write manual SQL statements with op.execute for the migration? Or is there a better alternative?
Complete error log - https://bit.ly/3dOIXux

thoth/storages/graph/enums.py
fridex: > Sorry I didn't understand how this would affect the enumeration?

No need to list all the enums in the `ENUM()`. If we add another one, it will be automatically populated there.

thoth/storages/data/alembic/versions/76413cbe99bf_added_source_type_to_adviser_run_table.py
fridex: > Should we have just write manual SQL statements with op.execute for the migration? Or is there a better alternative?

Check migrations we already have. I don't recall if we did something similar to this (to learn from). Othrewise, raw  SQL statements will be necessary I guess.

saisankargochhayat: @fridex Thanks, I have added a manual migration, PTAL. :smile:

saisankargochhayat:
I have added the `normalize_os_version` method in this pull request.

saisankargochhayat:
> The code looks good to me. Were all the related issues resolved?

image
The local alembic version seem to go all right, I tried one or two get queries also, anything else I should check for?

PR #1879, opened 2020-07-07
Introduce method for retrieving inspection ids and count

fridex:

- [ ] Yes
- [x] No


pacospace:
@fridex PTAL so we can release new version :)

PR #1881, opened 2020-07-07
Adjust SI classes

pacospace:

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Modify SI classes to have a hierarchy.

### DIFF
--- a/thoth/storages/security_indicators.py
+++ b/thoth/storages/security_indicators.py
@@ -67,11 +67,11 @@ class SIBanditStore(_SecurityIndicatorBase):
         self.ceph = CephStore(prefix=prefix)
         self.security_indicator_id = security_indicator_id
 
-    def retrieve_si_bandit_document(self) -> Dict[str, Any]:
+    def retrieve_document(self) -> Dict[str, Any]:
         """Retrieve SI bandit document."""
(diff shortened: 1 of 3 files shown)


thoth/storages/security_indicators.py
```diff
@@ -59,45 +103,63 @@ def si_aggregated_document_exists(self) -> bool:
...
-        return self.ceph.get_document_listing()
-
-    def get_document_count(self) -> int:
-        """Get number of documents present."""
-        return len(tuple(self.get_listing()))/3
+        self.bandit.check_connection()
+        self.cloc.check_connection()
+        self.aggregated.check_connection()
+
+    def exists(self, si_type: str) -> bool:
```
fridex, line 135: Not sure about usage of this, it can be substituted with a more type-safe version:

```python
store = SecurityIndicatorsResultsStore("some-id")
store.connect()
store.si_aggregated_document_exists()
```

for the other two (bandit, cloc) respective methods could be introduced in used adapters.

fridex: ... and maybe we should completely discard name of the target from the method so it gets abstracted away. That way we can use adapters without any notion which results they encapsulate (and inherit from a common interface).

PR #1882, opened 2020-07-07
Create queries for SI retrieval for certain package name, version , index

pacospace:

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Add datetime to table SI aggregated to allow filtering.
Create queries to retrieve SI for package name, version, index.


thoth/storages/graph/postgres.py
```diff
@@ -2125,6 +2125,117 @@ def get_analyzer_documents_count_all(self) -> int:
...
+            query = query.filter(PythonPackageVersionEntity.package_version == package_version)
+
+        if index_url is not None:
+            query = query.filter(PythonPackageIndex.url == index_url)
+        
+        query = query.with_entities(
+            SecurityIndicatorAggregatedRun
+        )
+
+        conditions = [
```
fridex, line 2155: I think they can be put directly to filter.

```python
    return query.filter(
            SIAggregated.python_package_version_entity_id == SecurityIndicatorAggregatedRun.id,
            PythonPackageVersionEntity.python_package_index_id == PythonPackageIndex.id,
    )
```

thoth/storages/graph/postgres.py
```diff
@@ -2125,6 +2125,117 @@ def get_analyzer_documents_count_all(self) -> int:
...
+        """
+        with self._session_scope() as session:
+            query = self._construct_query_get_si_aggregated_python_package_version(
+                session=session,
+                package_name=package_name,
+                package_version=package_version,
+                index_url=index_url
+            )
+
+            result = query.order_by(SecurityIndicatorAggregatedRun.datetime.desc()).first()
```
fridex, line 2233: The ordering can be an expensive operation. Do we need it here? Does `first` obtain the very last inserted? Or compare based on reverted ids?

pacospace: in theory we can have si for same package run multiple times and we want to retrieve only the latest one. The alternative is that we do an upsert if there is a more recent datetime and we update the date, so when we use the query, we should receive only one result per package name version and we do not need to sort. 

Regarding `first()` I think it will just give the first result from the query, so we are not sure is the latest inserted I guess.

fridex: The thing you described is relevant also for other entries (e.g. solver queries, ...). We can run the same workflow multiple times and obtain results from the last one (in most cases).

OK, let's have this in and fine-tune perf later as needed.

thoth/storages/graph/postgres.py
```diff
@@ -2125,6 +2125,115 @@ def get_analyzer_documents_count_all(self) -> int:
         with self._session_scope() as session:
             return session.query(PackageExtractRun).distinct(PackageExtractRun.analysis_document_id).count()
 
+    def _construct_query_get_si_aggregated_python_package_version(
+        self,
+        session: Session,
+        package_name: str,
+        package_version: str,
+        index_url: str,
```
fridex, line 2133:
```python
        package_name: Optional[str],
        package_version: Optional[str],
        index_url: Optional[str],
```

Issue #1871, opened 2020-07-01
pypi.python.org is pypi.org


fridex:
**Describe the bug**

Some stack configurations use pypi.python.org as a source of packages. This old URL redirects to pypi.org. We should add capability to automatically remap all the index URLs pointing to legacy pypi.python.org to pypi.org.

As of now, we maintain both entries as separate Python package indexes. This, in fact, doubles the number of packages stored.

**Expected behavior**

Remap all pypi.python.org URLs to pypi.org.

This change should go also to thoth-python.

saisankargochhayat:
From the schema, it seems `PythonPackageVersionEntity`, `PythonPackageVersion`, `PythonPackageRequirement`, depends on `PythonPackageIndex`.

goern:
is this still valid?

fridex:
> is this still valid?

Not anymore.

PR #1897, opened 2020-07-10
Map old domain to new

saisankargochhayat:
Related to #1871 
We map all new requests to `_get_or_create_python_package_index` to the new domain

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -5414,7 +5414,6 @@ class GraphDatabase(SQLBase):
         package_version = self.normalize_python_package_version(package_version)
         index_url = GraphDatabase.normalize_python_index_url(index_url)
 
-
         with self._session_scope() as session:
             query = (
                 session.query(PythonPackageVersionEntity)
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -371,6 +371,14 @@ class GraphDatabase(SQLBase):
             return "rhel"
 
         return os_name
+    
+    @staticmethod
+    def normalize_python_index_url(index_url: Optional[str]) -> Optional[str]:
+        """Map python index url."""
+        if index_url == "https://pypi.python.org/simple":
+            return "https://pypi.org/simple"
+
+        return index_url
 
     def get_analysis_metadata(self, analysis_document_id: str) -> Dict[str, Any]:
         """Get metadata stored for the given analysis document.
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -4616,7 +4624,7 @@ def _get_or_create_python_package_index(
         if python_package_index is None:
             if only_if_enabled:
                 raise PythonIndexNotRegistered(f"Python package index {index_url!r} is not know to system")
-
+            index_url = GraphDatabase.map_python_index_url(index_url)
```
pacospace, line 4627: Why not at the beginning of the method? :)

saisankargochhayat: Ah right! Changed it :)

thoth/storages/graph/postgres.py
```diff
@@ -371,6 +371,14 @@ def map_os_name(os_name: Optional[str]) -> Optional[str]:
             return "rhel"
 
         return os_name
+    
+    @staticmethod
+    def map_python_index_url(index_url: Optional[str]) -> Optional[str]:
```
fridex, line 376: Can this be renamed to `normalize_python_index_url` so it matches other normalize_* methods?

thoth/storages/graph/postgres.py
```diff
@@ -4611,12 +4619,12 @@ def _get_or_create_python_package_index(
         session: Session, index_url: str, only_if_enabled: bool = True
     ) -> Optional[PythonPackageIndex]:
         """Get or create Python package index entry with a check the given index is enabled."""
+        index_url = GraphDatabase.map_python_index_url(index_url)
```
fridex, line 4622: Nice, we can do this for all the functions in this `postgres.py` module that accept index_url.

thoth/storages/graph/postgres.py
saisankargochhayat: Thanks!

thoth/storages/graph/postgres.py
saisankargochhayat: Added it in all functions having that parameter in `postgres.py`. Btw don't all of them ultimately depend on - `_get_or_create_python_package_index`?

thoth/storages/graph/postgres.py
```diff
@@ -5361,6 +5412,8 @@ def get_python_package_required_symbols(self, package_name: str, package_version
         """Get required symbols for a Python package in a specified version."""
         package_name = self.normalize_python_package_name(package_name)
         package_version = self.normalize_python_package_version(package_version)
+        index_url = GraphDatabase.normalize_python_index_url(index_url)
+
```
fridex, line 5416: Unnecessary space?

saisankargochhayat: thanks removed

PR #1929, opened 2020-08-12
add cache and return None if empty

KPostOffice:

https://github.com/thoth-station/adviser/pull/1108

## Description

Add cache to getting SI results as well as returning None if it is not set.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -177,6 +177,7 @@ _GET_PYTHON_PACKAGE_REQUIRED_SYMBOLS_CACHE_SIZE = int(
     os.getenv("THOTH_STORAGE_GET_PYTHON_PACKAGE_REQUIRED_SYMBOLS_CACHE_SIZE", 4096)
 )
 _GET_PYTHON_ENVIRONMENT_MARKER_CACHE_SIZE = int(os.getenv("THOTH_GET_PYTHON_ENVIRONMENT_MARKER_CACHE_SIZE", 4096))
+_GET_SI_AGGREGATED_PYTHON_PACKAGE_VERSION_CACHE_SIZE = int(os.getenv("THOTH_GET_PYTHON_ENVIRONMENT_MARKER_CACHE_SIZE", 4096))
 
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -2254,13 +2256,16 @@ def get_si_aggregated_python_package_version(
         """
         index_url = GraphDatabase.normalize_python_index_url(index_url)
         with self._session_scope() as session:
-            query = self._construct_query_get_si_aggregated_python_package_version(
+            query = self._construct_query_GET_SI_AGGREGATED_PYTHON_PACKAGE_VERSION(
```
saisankargochhayat, line 2259: Would using the capitalization work? The function name seems to be lowercase - https://github.com/thoth-station/storages/blob/e2ad480759750823cb67be9ba57f2e006169c42f/thoth/storages/graph/postgres.py#L2160

KPostOffice: Oh, I must have done this accidentally when I capitalized the cache size...

thoth/storages/graph/postgres.py
```diff
@@ -2199,7 +2200,7 @@ def si_aggregated_python_package_version_exists(
         """Check if Aggregate Security Indicators (SI) results exists for Python package version."""
         index_url = GraphDatabase.normalize_python_index_url(index_url)
         with self._session_scope() as session:
-            query = self._construct_query_get_si_aggregated_python_package_version(
+            query = self._construct_query_GET_SI_AGGREGATED_PYTHON_PACKAGE_VERSION(
```
saisankargochhayat, line 2203: This also :smile:

thoth/storages/graph/postgres.py
```diff
@@ -2261,6 +2263,9 @@ def get_si_aggregated_python_package_version(
                 index_url=index_url
             )
 
+            if query.count() == 0:
```
fridex, line 2266: I think this will introduce two queries - one for count and the other for actual data retrieval. It should be more optimal to move this check down bellow:

```python
result = query.order_by(SecurityIndicatorAggregatedRun.datetime.desc()).first()
if not result:
    return None

# ... result.to_dict() logic
```

Issue #1948, opened 2020-08-20
Logging error when checking database schema version


fridex:
**Describe the bug**

When checking the database schema, I obtain this logging error:

```
--- Logging error ---
Traceback (most recent call last):
  File "/usr/lib64/python3.6/logging/__init__.py", line 994, in emit
    msg = self.format(record)
  File "/usr/lib64/python3.6/logging/__init__.py", line 840, in format
    return fmt.format(record)
  File "/usr/lib64/python3.6/logging/__init__.py", line 577, in format
    record.message = record.getMessage()
  File "/usr/lib64/python3.6/logging/__init__.py", line 338, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Call stack:
  File "<stdin>", line 1, in <module>
  File "/home/fpokorny/git/thoth-station/storages/thoth/storages/graph/postgres.py", line 294, in connect
    if not self.is_schema_up2date():
  File "/home/fpokorny/git/thoth-station/storages/thoth/storages/graph/postgres.py", line 362, in is_schema_up2date
    database_heads,
Message: 'The database schema is not in sync with library revisions, the current library revision heads: %r, database heads'
Arguments: ({'83e2900c3721'}, {'7c261487d4d4'})
```

**To Reproduce**
```python
from thoth.storages import GraphDatabase
graph = GraphDatabase()
graph.connect()
graph.is_schema_up2date()
# see the error
```

PR #1949, opened 2020-08-20
Fix logging error when database schema is not up2date

fridex:

Fixes: #1948

- [x] Maybe


pacospace (approved):
:1st_place_medal:

PR #1963, opened 2020-08-24
Get si unsolved

### LINKED ISSUE #1947: Create a query to fetch all packages, that have been solved, but SI hasn't been run on it.
_Originally posted by @fridex in https://github.com/thoth-station/graph-refresh-job/issues/457#issuecomment-676128893_
Create and query in storages, that would look for all solved packages, and check if they exist in `si_aggregated_run` table and if it doesn't exist there, we should produce messages to schedule SI workflows.
- [ ] The table in which we have details of all the solved packages - 
- [ ] The SI table to check against is -

saisankargochhayat:
Closes #1947 
Related https://github.com/thoth-station/graph-refresh-job/issues/430
Added a query to check for all the packages that are solved by sovler, but SI workflow has not been run on them.


saisankargochhayat:
The `si_aggregated` table seems to be empty in the latest backup.

thoth/storages/graph/postgres.py
```diff
@@ -1880,6 +1880,78 @@ def get_analyzed_error_python_package_versions_count_all(
 
     # Unanalyzed Python Packages
 
+    def _construct_si_unanalyzed_python_package_versions_query(
+        self, session: Session, os_name: str = None, os_version: str = None, python_version: str = None, index_url: str = None
```
fridex, line 1884:
```suggestion
        self, session: Session, os_name: Optional[str] = None, os_version: Optional[str] = None, python_version: Optional[str] = None, index_url: Optional[str] = None
```

thoth/storages/graph/postgres.py
```diff
@@ -1880,6 +1880,78 @@ def get_analyzed_error_python_package_versions_count_all(
...
+                PythonPackageVersion.python_version == python_version)
+
+        # Finally filter these out.
+        query = query.filter(~exists().where(and_(*conditions)))
+
+        return query
+
+    def get_si_unanalyzed_python_package_versions_all(
+        self,
+        os_name: str = None,
```
fridex, line 1917:
```suggestion
        os_name: Optional[str] = None,
```

thoth/storages/graph/postgres.py
```diff
@@ -1880,6 +1880,78 @@ def get_analyzed_error_python_package_versions_count_all(
...
+        os_version: str = None,
```
fridex, line 1918:
```suggestion
        os_version: Optional[str] = None,
```

thoth/storages/graph/postgres.py
```diff
@@ -1880,6 +1880,78 @@ def get_analyzed_error_python_package_versions_count_all(
...
+        python_version: str = None,
```
fridex, line 1919:
```suggestion
        python_version: Optional[str] = None,
```

thoth/storages/graph/postgres.py
```diff
@@ -1880,6 +1880,78 @@ def get_analyzed_error_python_package_versions_count_all(
...
+        start_offset: int = 0,
+        count: Optional[int] = DEFAULT_COUNT,
+        distinct: bool = False,
```
fridex, line 1922: Do we want to have distinct by default?

```suggestion
        distinct: bool = True,
```

fridex (commented):
Just reading the query, it feels like it is correct :+1: (did not try it).

thoth/storages/graph/postgres.py
```diff
@@ -1880,6 +1880,78 @@ def get_analyzed_error_python_package_versions_count_all(
...
+            PythonPackageVersion.package_version.isnot(None),
+            PythonPackageIndex.url.isnot(None),
+            PythonPackageIndex.enabled.is_(True),
+        )
+
+        if index_url is not None:
+            query = query.filter(PythonPackageIndex.url == index_url)
+
+        # We find all rows that are same in PythonPackageVersion and SIAggregated table.
+        conditions = [PythonPackageVersion.entity_id == SIAggregated.id]
```
pacospace, line 1898: SIAggreated is not linked to PPV table, this condition would lead to wrong results: https://github.com/thoth-station/storages/blob/4a74226545c0a36dbca7ec248058b71905c94c6c/thoth/storages/graph/models.py#L1550
```
class SIAggregated(Base, BaseExtension):
    """A relation representing a Python package version entity analyzed by a si_aggregated_run."""

    __tablename__ = "si_aggregated"

    si_aggregated_run_id = Column(Integer, ForeignKey("si_aggregated_run.id", ondelete="CASCADE"), primary_key=True)
    python_package_version_entity_id = Column(
        Integer, ForeignKey("python_package_version_entity.id", ondelete="CASCADE"), primary_key=True
    )

    si_aggregated_run = relationship("SecurityIndicatorAggregatedRun", back_populates="python_package_version_entities")
    python_package_version_entity = relationship("PythonPackageVersionEntity", back_populates="si_aggregated_runs")
```

@fridex @saisankargochhayat

pacospace (commented):
One important comment!

thoth/storages/graph/postgres.py
saisankargochhayat: Do we want the same behavior in `get_unsolved_python_packages_version_all` ?

thoth/storages/graph/postgres.py
saisankargochhayat: Added an relation between both tables, please let me know if I am missing something? https://github.com/thoth-station/storages/pull/1963/commits/c6c14a5a8e3cc969df9001a5bcfe9109ba882d8d
Do I have to do an alembic upgrade here?

fridex: Yes, the migration will need to be created. Also, the syncing logic of SI results needs to be updated so that this relation is created in the database when syncing results.

CC @KPostOffice

thoth/storages/graph/postgres.py
```diff
@@ -1880,6 +1880,78 @@ def get_analyzed_error_python_package_versions_count_all(
...
+        randomize: bool = True,
+    ) -> List[Tuple[str, str]]:
+        """Retrieve solved Python package versions in Thoth Database, that are not anaylyzed by SI. 
+        Examples:
+        >>> from thoth.storages import GraphDatabase
+        >>> graph = GraphDatabase()
+        >>> graph.get_si_unanalyzed_python_package_versions_all()
+        [('crossbar', '0.10.0', 'https://pypi.org/simple'), ('tensorflow', '1.11.0', 'https://pypi.org/simple')]
+        """
+        os_version = OpenShift.normalize_os_version(os_name, os_version)
```
pacospace, line 1932: this line should go in `_construct_si_unanalyzed_python_package_versions_query`

thoth/storages/graph/postgres.py
pacospace: I don't think we should allow SI workflow to write on PPV directly, only solver should do that for consistency, as many methods depends on PPV, which is equivalent to say that it is solved.
If we introduce this migration and sync part, we are going to have issues, because PPV might be analyzed by SI and not by solver and this would create issues on adviser side as well.

pacospace: mmm actually after looking at when solver sync it creates also entity id (if not existing) and `conditions = [PythonPackageVersion.entity_id == SIAggregated.id]` is considering PPVE from PPV so we do not require PPV in PPVE and no migration. Sorry @saisankargochhayat for the confusion, I must have read too fast the other evening :(

fridex: Re consistency and issues on adviser side - The precondition for syncing SI workflows is the fact that solvers are already run (we have that already based on the system behavior) so no need to worry about these things.

thoth/storages/graph/postgres.py
saisankargochhayat: Sure!

saisankargochhayat:
@fridex @pacospace Thanks for the comments, I have added the comparison as we discussed and the alembic migration. We just need to add the relation while syncing.

pacospace (commented):
LGTM, syncing part is missing I think

saisankargochhayat:
> LGTM, syncing part is missing I think

Should I add that in this PR?

pacospace:
> > LGTM, syncing part is missing I think
> 
> Should I add that in this PR?

I think it make sense to be here. But you can have another PR sure.

I think that should work similar to adviser and inspection sync which uses this logic and fail if try to sync and a package is not solved: https://github.com/thoth-station/storages/blob/a87b6d3b79e04b001686968766bfaac4749cc260/thoth/storages/graph/postgres.py#L3728
For SI only part of that logic will be used, maybe a new method can be created

```
                    python_package_version = (
                        session.query(PythonPackageVersion)
                        .join(PythonPackageIndex)
                        .filter(PythonPackageVersion.package_name == package.name)
                        .filter(PythonPackageVersion.package_version == package.locked_version)
                        .filter(PythonPackageVersion.os_name == os_name)
                        .filter(PythonPackageVersion.os_version == os_version)
                        .filter(PythonPackageVersion.python_version == python_version)
                        .filter(PythonPackageIndex.url == index_url)
                        .first()
                    )

                # Check if we run the solver for a specific package already
                if python_package_version:
                    python_package_version = self._create_python_package_version(
                            session,
                            package_name=package_name,
                            package_version=package_version,
                            os_name=os_name,
                            os_version=os_version,
                            index_url=index_url,
                            python_package_metadata_id=python_package_version.python_package_metadata_id,
                            sync_only_entity=False,
                        )
                else:
                    raise SolverNotRun(
                        f"Trying to sync package {package_name!r} in version {package_version!r} "
                        f"not solved by solver-{os_name}-{os_version}-{python_version}"
                    )
```

the logic should modify https://github.com/thoth-station/storages/blob/a87b6d3b79e04b001686968766bfaac4749cc260/thoth/storages/graph/postgres.py#L4839

for the other part instead, once python_package_version
```
python_package_version_entity_id=python_package_version.entity_id
python_package_version_id = python_package_version.id
```

and https://github.com/thoth-station/storages/blob/a87b6d3b79e04b001686968766bfaac4749cc260/thoth/storages/graph/postgres.py#L4877 to have ids from PPV and PPVE

WDYT @fridex ?

saisankargochhayat:
https://github.com/thoth-station/storages/pull/1963/commits/757830f6eec1c878cc1806fb5c5cbe2949457f46
Would this work? Assuming the python package would already exist in `PythonPackageVersion`  if a SI run is being scheduled? @pacospace

fridex:
> WDYT @fridex ?

+1

fridex:
> [757830f](https://github.com/thoth-station/storages/commit/757830f6eec1c878cc1806fb5c5cbe2949457f46)
> Would this work? Assuming the python package would already exist in `PythonPackageVersion` if a SI run is being scheduled? @pacospace

Briefly checking, it looks like you are missing the check described by @pacospace.

pacospace:
> [757830f](https://github.com/thoth-station/storages/commit/757830f6eec1c878cc1806fb5c5cbe2949457f46)
> Would this work? Assuming the python package would already exist in `PythonPackageVersion` if a SI run is being scheduled? @pacospace

Thanks @saisankargochhayat  @fridex Yes it is currently missing that check.

As of now, this would write directly to PPV table which is what we do not want, if we run SI and solver at the same time (which can happen currently) for the same package and SI finishes first it will write to PPV. 

We want only solver to write directly to PPV. Other components like adviser and inspection when syncing check if that package has been solved first otherwise the sync fails. It's to guarantee consistency. 

Imagine if SI sync first in PPV, as PPV == solved packages, some other components could use that info, but it would not be True. As I explained before, we should have a similar check as for adviser and inspection during sync to avoid inconsistencies.

thoth/storages/graph/postgres.py
```diff
@@ -4831,54 +4944,69 @@ def sync_security_indicator_aggregated_result(self, document: dict) -> None:
...
-            SIAggregated.get_or_create(
-                session,
-                si_aggregated_run_id=si_aggregated_run.id,
-                python_package_version_entity_id=python_package_version_entity.id,
-            )
+                python_package_version, _ = PythonPackageVersion.get_or_create(
+                    session,
+                    package_name=package_name,
+                    package_version=package_version,
+                    python_package_index_id=python_package_index.id,
```
pacospace, line 4970: This component don't have to create entries in PPV or PPVE. The method above should return PPVE ID and PPV ID to be passed to:

```
SIAggregated.get_or_create(
                    session,
                    si_aggregated_run_id=si_aggregated_run.id,
                    python_package_version_entity_id=python_package_version_entity.id,
                    python_package_version=python_package_version.id,
                )
```

If the method above does not return those required values it will fail. If some package is not solved for some reason, we cannot sync the corresponding package in the database for consistency. The system should heal itself automatically rescheduling eventually an SI.

thoth/storages/graph/postgres.py
```diff
@@ -4822,6 +4894,47 @@ def sync_revsolver_result(self, document: Dict[str, Any]) -> None:
                     platform=entry["platform"],
                 )
 
+    def _check_package_solved(self, session: Session, package_name: str, package_version: str,
```
pacospace, line 4897: This method should return package_version_entiy ID and package_version ID required below. If those values are not returned the method will fail, which is what we want. If some package is not solved first, we do not want to sync anything else in the database for that package.

pacospace (commented):
Two comments on the logic, otherwise we are almost done :) Thanks!!

thoth/storages/graph/postgres.py
pacospace: > This method should return package_version_entiy ID and package_version ID required below in the SIAggregated sync. If those values are not returned the method will fail, which is what we want. If some package is not solved first, we do not want to sync anything else in the database for that package.

saisankargochhayat: https://github.com/thoth-station/storages/pull/1963/commits/f409f97503d2f429e6b55ba3f072dc932ae4393e
Something like this?

thoth/storages/graph/postgres.py
```diff
@@ -4822,6 +4894,57 @@ def sync_revsolver_result(self, document: Dict[str, Any]) -> None:
...
+                              package_index: str, os_name: str, os_version: str, python_version: str) -> Tuple:
```
pacospace, line 4898: we do not require solver (os_name, os_version, python_version), we just want to know if it exists from any solver, because SI is not related to solver.

thoth/storages/graph/postgres.py
```diff
@@ -4822,6 +4894,57 @@ def sync_revsolver_result(self, document: Dict[str, Any]) -> None:
...
+        # Check if the package has a known index
+        if package_index:
+            python_package_version = (
+                session.query(PythonPackageVersion)
+                .join(PythonPackageIndex)
+                .filter(PythonPackageVersion.package_name == package_name)
+                .filter(PythonPackageVersion.package_version == package_version)
+                .filter(PythonPackageVersion.os_name == os_name)
+                .filter(PythonPackageVersion.os_version == os_version)
+                .filter(PythonPackageVersion.python_version == python_version)
```
pacospace, line 4909: This is not required also :) See comment above

pacospace: This should be substituted  with just a check if package_version, version, index and return the record otherwise fail.

thoth/storages/graph/postgres.py
pacospace: Yes but realized we do not need solver info here. See comments below

pacospace: We just need to know if the package version index exist in the database in PPV table, which means it is solved. Otherwise sync fails.

thoth/storages/graph/postgres.py
```diff
@@ -4822,6 +4894,57 @@ def sync_revsolver_result(self, document: Dict[str, Any]) -> None:
...
+        # We return python_package_entity id and python_package_version id if solver has been run. 
+        python_package_index = self._get_or_create_python_package_index(
+                    session, index_url=package_index, only_if_enabled=True
+                )
+        python_package_version_entity, _ = PythonPackageVersionEntity.get_or_create(
+            session,
+            package_name=package_name,
+            package_version=package_version,
+            python_package_index_id=python_package_index.id,
+        )
```
pacospace, line 4944: Not required, we don't create anything in SI sync beside record in SI table and relation table :)

thoth/storages/graph/postgres.py
```diff
@@ -4822,6 +4894,57 @@ def sync_revsolver_result(self, document: Dict[str, Any]) -> None:
...
+
+        return python_package_version_entity.id, python_package_version.id
```
pacospace, line 4946: `python_package_version_entity.id` -> `python_package_version.entity_id` we take both info from PPV table

saisankargochhayat:
@pacospace thanks for all the suggestions, we now get both entity id and ppv id from the same table :+1:

thoth/storages/graph/postgres.py
saisankargochhayat: Thanks I have removed it

thoth/storages/graph/postgres.py
```diff
@@ -4822,6 +4894,41 @@ def sync_revsolver_result(self, document: Dict[str, Any]) -> None:
...
+                index_url=package_index,
+                sync_only_entity=False,
+            )
+        else:
+            raise SolverNotRun(
+                f"Trying to sync package {package_name!r} in version {package_version!r} "
+                f"not solved by any solver."
+            )
+        # We return python_package_entity id and python_package_version id if solver has been run.
+        return python_package_version.entity_id, python_package_version.id
```
pacospace, line 4930:
```
            python_package_version = (
                session.query(PythonPackageVersion)
                .join(PythonPackageIndex)
                .filter(PythonPackageVersion.package_name == package_name)
                .filter(PythonPackageVersion.package_version == package_version)
                .filter(PythonPackageIndex.url == package_index if package_index else None)
                .first()
            )

        # Check if we run any solver for a specific package already
        if not python_package_version:
            raise SolverNotRun(
                f"Trying to sync package {package_name!r} in version {package_version!r} "
                f"not solved by any solver."
            )

        # We return python_package_entity id and python_package_version id if any solver has been run.
        return python_package_version.entity_id, python_package_version.id
```

pacospace: No need to check index either, as we schedule solver only for known index. We just want to know if that package has been solved from any solver, one is enough :)

saisankargochhayat: Thanks changed it :+1:

thoth/storages/graph/postgres.py
```diff
@@ -1880,6 +1880,78 @@ def get_analyzed_error_python_package_versions_count_all(
...
+
+        if os_name:
+            conditions.append(PythonPackageVersion.os_name == os_name)
+
+        if os_version:
+            conditions.append(PythonPackageVersion.os_version == os_version)
+
+        if python_version:
+            conditions.append(
+                PythonPackageVersion.python_version == python_version)
```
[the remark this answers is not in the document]
saisankargochhayat: Thanks!

pacospace (approved):
Nice, It looks all good. I didn't test queries and sync but I'm sure you did :D !! @saisankargochhayat

harshad16:
@saisankargochhayat  can you please rebase

saisankargochhayat:
> @saisankargochhayat can you please rebase

Done :+1:

PR #1964, opened 2020-08-26
query.all returns type List[result] not List[str]

KPostOffice:
https://github.com/thoth-station/package-releases-job/issues/440

`query.all()` return type is `List[result]` not `List[str]`.  `result` can be treated as a `tuple`. So because `query.all()` returns:
```
[("foo",), ("bar",), ("baz",)]
```
for example. We can use list comprehension to turn this into ["foo", "bar", "baz"] to match the type hint return type.  This function is only used in package-releases-job rn so this change will not cause any breakages.  It will fix the bug stated above.


pacospace (approved):
Thanks!! :100:

fridex (approved):
I think we need one more approval.

Issue #2007, opened 2020-09-17
SI unanalyzed and SI analyzed queries produce wrong results.


pacospace:
**Describe the bug**
SI unalyzed and analyzed queries produce wrong results.

**To Reproduce**
1. have a look at Grafana

**Expected behavior**

**Screenshots**
After SI Argo workflow analyzing one package, this is the result in data:

![Screenshot from 2020-09-17 10-17-36](https://user-images.githubusercontent.com/27498679/93491377-e928ab80-f909-11ea-9ec2-b11c727f3400.png)

- SI unanalyzed should go down by 1 not up 
- SI analyzed should go up by 1 not 50k.

**Additional context**

pacospace:
I found the issue, I will take care :)

pacospace:
related-to: https://github.com/thoth-station/adviser/issues/1178

PR #2013, opened 2020-09-21
Adjust SI queries

pacospace:

Fixes: https://github.com/thoth-station/storages/issues/2007
Related-To: https://github.com/thoth-station/adviser/issues/1178
Related-To: https://github.com/thoth-station/thoth-application/issues/423

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

It affects:

- [ ] adviser
- [ ] investigator
- [ ] graph-refresh-job
- [ ] metrics-exporter

that requires new releases after Thoth storages to have SI feature ready.

- Adjust queries due to change in SI schema.
- Introduce query to retrieve packages that has SI information

```
SI analyzed query count: 2
SI analyzed query all: [('fbprophet', '0.4', 'https://pypi.org/simple'), ('rsa', '4.0', 'https://pypi.org/simple')]
SI unanalyzed query count: 36735
Python packages releases 36737
```

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -1592,42 +1592,57 @@ class GraphDatabase(SQLBase):
     # SI Analyzed Python Packages
 
     def _construct_si_analyzed_python_package_versions_query(
-        self, session: Session, index_url: Optional[str] = None
+        self, session: Session
     ) -> Query:
         """Construct query for packages analyzed by solver and analyzed by SI."""
-        index_url = GraphDatabase.normalize_python_index_url(index_url)
-        query = session.query(PythonPackageVersion).filter(
-            PythonPackageVersion.package_version.isnot(None),
-            PythonPackageIndex.url.isnot(None),
-            PythonPackageIndex.enabled.is_(True),
-        )
-
-        if index_url is not None:
-            query = query.filter(PythonPackageIndex.url == index_url)
+        query = session.query(SecurityIndicatorAggregatedRun)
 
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -1651,7 +1672,7 @@ def _construct_si_unanalyzed_python_package_versions_query(
             query = query.filter(PythonPackageIndex.url == index_url)
 
         # We find all rows that are same in PythonPackageVersion and SIAggregated table.
-        conditions = [PythonPackageVersion.id == SIAggregated.python_package_version_id]
+        conditions = [PythonPackageVersion.entity_id == SIAggregated.python_package_version_entity_id]
```
saisankargochhayat, line 1675: @pacospace could you please explain why we switched back to comparing entity_id?

pacospace: Because for each `entity_id` exist `number of solvers x python_package_version_id` therefore when we query for PPV and SI we always receive still the same python_package_version because it exist for different solvers, that is why the number fo SI Unanalyzed was not going down, unless we look for PPVE record for the PPV, because that is one.

saisankargochhayat: Makes sense, thanks!

pacospace: If I'm not clear I can draw something, let me know :)

saisankargochhayat (approved):
LGTM, thanks :100:

PR #2021, opened 2020-09-28
add flag for missing src distro

KPostOffice:

https://github.com/thoth-station/thoth-application/issues/507

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

We need to have a flag indicating whether or not we can get security indicator data

## Description

This flag will be updated by investigator using a new message which can be sent to update the value of a single row


pacospace (approved):
:100:

PR #2033, opened 2020-09-29
change flag name and alembic commands

KPostOffice:

https://github.com/thoth-station/thoth-application/issues/507

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Change name of flag and also change how column name is updated, use update column instead of deleting and remaking


fridex (approved):
:+1:

Issue #2038, opened 2020-09-30
Extend kebechet github installation table with software stack ID for Internal Trigger feature [5pt]


pacospace:
As Thoth.

I would like to provide my services to users of Kebechet and keep track of what managers are used by users so that we can prioritize future Kebechet features and provide Kebechet services faster.

BLOCKERS: 
- [x] https://github.com/thoth-station/storages/issues/1675

Tasks:

- [x] Extend kebechet_github_installation table to have one column for software stack ID 
- [x] Create migration.
- [x] Create new method to query for current state of managers used by user.

References:
- https://github.com/thoth-station/kebechet/issues/571

Issue #1675, opened 2020-03-25
PythonSoftwareStack table has three records!!


pacospace:
**Describe the bug**
`PythonSoftwareStack` currently has three columns: `software_stack_type`, `performance_score` and `overall_score`. This table is linked to `PythonRequirementsLocked` and `ExternalPythonRequirementsLocked` which are respectively linked to `PythonPackageVersion` and `PythonPackageVersionEntity`. Every service that sync a software stack (Adviser, Amun, etc..) is currently referring to the same record in `PythonSoftwareStack`. This cause an issue if we want to retrieve the software stack used by a User or suggested by Adviser querying the database because there is no unique `PythonSoftwareStack` record created each time.

**To Reproduce**
1. Go to PythonSoftwareStack table in the database. There are only three entries.
Other way is to create a query to retrieve python package version or python package version entities for a specific python software stack (PythonSoftwareStack table).

**Expected behavior**
Every packages in a software stacks is identified by a unique software stack id. One solution could be to create a unique hash and add this column to PythonSoftwareStack, and only modify the method https://github.com/thoth-station/storages/blob/d1bc3db14fe50647f08606c50940f876ed665134/thoth/storages/graph/postgres.py#L3560.

```
from
        software_stack, _ = PythonSoftwareStack.get_or_create(
            session,
            performance_score=performance_score,
            overall_score=overall_score,
            software_stack_type=software_stack_type,
        )
to
        software_stack, _ = PythonSoftwareStack.get_or_create(
            session,
            `hash=_create_unique_hash(requirements_lock=requirements_lock)`
            performance_score=performance_score,
            overall_score=overall_score,
            software_stack_type=software_stack_type,
        )
```
with new method to create the hash.

fridex:
> ```
> from
>         software_stack, _ = PythonSoftwareStack.get_or_create(
>             session,
>             performance_score=performance_score,
>             overall_score=overall_score,
>             software_stack_type=software_stack_type,
>         )
> to
>         software_stack, _ = PythonSoftwareStack.get_or_create(
>             session,
>             `hash=_create_unique_hash(requirements_lock=requirements_lock)`
>             performance_score=performance_score,
>             overall_score=overall_score,
>             software_stack_type=software_stack_type,
>         )
> ```


Other solutions as discussed on the call:

[...]

PR #2046, opened 2020-10-07
Changes in schema for bug, internal trigger and data analysis

pacospace:

Fixes: https://github.com/thoth-station/storages/issues/1675
Fixes: https://github.com/thoth-station/storages/issues/2047
Fixes: https://github.com/thoth-station/storages/issues/2038
Fixes: https://github.com/thoth-station/s2i-thoth/issues/86
Fixes: https://github.com/thoth-station/kebechet/issues/573
Related-To: https://github.com/thoth-station/kebechet/issues/571
Related-To: https://github.com/thoth-station/workflow-helpers/issues/148

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

- Modify schema to solve issue and allow query required
- Add alembic migration
- Introduce queries


saisankargochhayat:
Do we commit the schema.png?

fridex (commented):
Would it make sense to split PR to multiple ones so it's easier to review? This one is fixing 3 issues at once.

thoth/storages/graph/postgres.py
```diff
@@ -3574,10 +3711,21 @@ def _create_python_package_version(
 
         return python_package_version
 
+    @staticmethod
+    def _create_hash(inputs: List[int]) -> str:
+        """Create hash string using ssdeep.
+
+        Reference: https://python-ssdeep.readthedocs.io/en/latest/installation.html#id11
+        """
+        file_ids = "-".join([str(ppr) for ppr in inputs])
```
fridex, line 3720: As hash operates on the binary data, it might make sense to keep items as integers rather than converting them to strings.

pacospace: but `ssdeep.hash()` expects a string not a list, how do we pass integers there?

Moreover using only integers sorted might give us the following problem:

- software stack with id 1 , 3 , 4 would results in 134
- software stack with id 1 , 34 would results in 134

and this is not correct, that is why I added the `-` when creating the string, so we don't fall into this bug.

pacospace: Or we could hash all ids and merge them in a single string, but it would be the same I guess, just more expensive

pacospace:
> Would it make sense to split PR to multiple ones so it's easier to review? This one is fixing 3 issues at once.

I removed many things, there are three commits only now:

- add thoth-ssdeep
- modify models and create migrations
- modify sync and create queries

thoth/storages/graph/postgres.py
fridex: > but `ssdeep.hash()` expects a string not a list of int or int, how do we pass integers there?

Hashes work on binary data.

> Moreover using only integers sorted might give us the following problem:
> 
> * software stack with id 1 , 3 , 4 would results in 134
> * software stack with id 1 , 34 would results in 134
> and this is not correct, that is why I added the `-` when creating the string, so we don't fall into this bug.

The issue you described is occurring in the string implementation. As hashes work on binary data, having allocated a fixed size for integers representing packages will yield expected results. The hash itself has no notion about `-` delimiter and the comparison will semantically not work. See comparison using [1].

[1] https://python-ssdeep.readthedocs.io/en/latest/api.html#ssdeep.compare

pacospace: Thanks @fridex 

I tested with

```
import ssdeep
INPUTS_1 = [1, 3, 4]
INPUTS_2 = [1, 34]

hashes = []
for file_ in [INPUTS_1, INPUTS_2]:

    hash_ = ssdeep.hash(bytes(file_))
    hashes.append([file_, hash_])

print(hashes)
print(ssdeep.compare(hashes[0][1], hashes[1][1]))
```

and the result was:
```
[[[1, 3, 4], '3:5:5'], [[1, 34], '3:On:On']]
0
```

So we will pass a list of integers in bytes correct?

pacospace: that does not work for number above 256 :/ what about this libarary https://pypi.org/project/hashids/ @fridex @saisankargochhayat ?

saisankargochhayat: Seems popular and might work well, but it's not a fuzzy hash if that's something we require.

pacospace: > Seems popular and might work well, but it's not a fuzzy hash if that's something we require.

Right :/

pacospace: hashids allow to pass different ids actually: `hashids.encode(*table_ids)`

thoth/storages/graph/postgres.py
```diff
@@ -3574,10 +3704,25 @@ def _create_python_package_version(
...
+    @staticmethod
+    def _create_fuzzy_hash(sorted_ids: List[int]) -> str:
+        """Create fuzzy hash from sorted list of integers.
+
+        Reference: https://pypi.org/project/python-ssdeep/
+        Reference: https://docs.python.org/3/library/stdtypes.html#int.to_bytes
+        """
+        hash_ = ""
+        for object_id in sorted_ids:
+            # lenght of bytes required for the integer
```
saisankargochhayat, line 3716: str(length/lenght)

pacospace: what is it? :D

saisankargochhayat: # (lenght -> length) of bytes required for the integer :stuck_out_tongue:

pacospace: thanks sir :D I didn't see the typo actually :rocket:

goern:
@fridex @pacospace what needs to be done to merge this?

pacospace:
> @fridex @pacospace what needs to be done to merge this?

No more changes from my side. It should be ready to go, tested all cases and queries. Also the method for fuzzy hash creation as discussed with @fridex has been added:

```
    @staticmethod
    def _create_fuzzy_hash(sorted_ids: List[int]) -> str:
        """Create fuzzy hash from sorted list of integers.

        Reference: https://pypi.org/project/python-ssdeep/
        Reference: https://docs.python.org/3/library/stdtypes.html#int.to_bytes
        """
        hash_ = ""
        for object_id in sorted_ids:
            # lenght of bytes required for the integer
            length = (object_id.bit_length() + 7) // 8
            # no negative ids are passed
            hash_ += ssdeep.hash(object_id.to_bytes(length=length, byteorder='big'))
        return hash_
```

pacospace:
> > @fridex @pacospace what needs to be done to merge this?
> 
> No more changes from my side. It should be ready to go, tested all cases and queries. Also the method for fuzzy hash creation as discussed with @fridex has been added:
> 
> ```
>     @staticmethod
>     def _create_fuzzy_hash(sorted_ids: List[int]) -> str:
>         """Create fuzzy hash from sorted list of integers.
> 
>         Reference: https://pypi.org/project/python-ssdeep/
>         Reference: https://docs.python.org/3/library/stdtypes.html#int.to_bytes
>         """
>         hash_ = ""
>         for object_id in sorted_ids:
>             # lenght of bytes required for the integer
>             length = (object_id.bit_length() + 7) // 8
>             # no negative ids are passed
>             hash_ += ssdeep.hash(object_id.to_bytes(length=length, byteorder='big'))
>         return hash_
> ```

Hey @fridex anything more here? :) Can we merge it?

goern:
@fridex is this good to merge? it seems to block https://github.com/thoth-station/workflow-helpers/pull/154

thoth/storages/graph/models.py
```diff
@@ -933,13 +910,45 @@ class PythonSoftwareStack(Base, BaseExtension):
...
     python_package_requirements = relationship("PythonRequirements", back_populates="python_software_stack")
-    python_package_versions = relationship("PythonRequirementsLock", back_populates="python_software_stack")
-    python_package_versions_entities = relationship(
-        "ExternalPythonRequirementsLock", back_populates="python_software_stack"
+    python_package_requirements_locked = relationship("PythonRequirementsLock", back_populates="python_software_stack")
+
+    advised_by = relationship("Advised", back_populates="python_software_stack")
+    inspection_runs = relationship("InspectionRun", back_populates="inspection_software_stack")
+
+    kebechet_github_installation = relationship("KebechetGithubAppInstallations", back_populates="advised_software_stack")
```
fridex, line 929: :+1:

thoth/storages/graph/postgres.py
```diff
@@ -3599,42 +3744,101 @@ def _create_python_software_stack(
...
-            overall_score=overall_score,
-            software_stack_type=software_stack_type,
-        )
-
         if requirements is not None:
             python_package_requirements = self._create_python_package_requirement(session, requirements)
-            for python_package_requirement in python_package_requirements:
-                PythonRequirements.get_or_create(
+            # Create unique hash for requirements to go into PythonRequirements
+            requirements_ids = [int(ppr.id) for ppr in python_package_requirements]
```
fridex, line 3750: You can do this with in-place sorting:
```python
            requirements_ids = [int(ppr.id) for ppr in python_package_requirements].sorted()
```

fridex (commented):
Very nice! Based on the schema generated, it looks like `HasPythonRequirementsLock` is an orphan - it links to just `python_package_version`

pacospace:
> HasPythonRequirementsLock

It links PythonRequirementsLock and PythonPackageVersion, did I miss something?

```
class HasPythonRequirementsLock(Base, BaseExtension):
    """The requirement from Pipfile.lock."""

    __tablename__ = "has_python_requirement_lock"

    python_requirements_lock_id = Column(
        Integer, ForeignKey("python_requirements_lock.id", ondelete="CASCADE"), primary_key=True
    )
    python_package_version_id = Column(
        Integer, ForeignKey("python_package_version.id", ondelete="CASCADE"), primary_key=True
    )

    python_requirements_lock = relationship("PythonRequirementsLock", back_populates="python_requirements_lock")
    python_package_version = relationship(
        "PythonPackageVersion", back_populates="python_software_stacks"
    )
```

fridex:
oh, bad upload above...

schema

It looks like an issue of the lib we use to generate the schema. Sources look good to me.

pacospace:
> oh, bad upload above...
> 
> schema
> 
> It looks like an issue of the lib we use to generate the schema. Sources look good to me.

can we merge it? :)

fridex:
/approve

> can we merge it? :)

ok, let's do it then...

fridex:
hm... I don't see my comment on the fixed bit length... how was it lost?! I will open an issue for that

pacospace:
> hm... I don't see my comment on the fixed bit length... how was it lost?! I will open an issue for that

I don't see comments either :(

harshad16:
super wierd the package - thoth-ssdeep is not being installed while the pytest run , i have tired to recreate the same scenario by installing the packages in same image, but it installs successfully. 
i m checking what is hindering the issue in the pipeline, it give out the same issue which we faced in https://github.com/thoth-station/s2i-thoth/issues/86

harshad16:
seems like the thoth-ssdeep fails in the image , would create a image in local and test out if we include the dev lib for fuzzy lib does the requirement satisfies, stay tuned for more updates.

pacospace:
> seems like the thoth-ssdeep fails in the image , would create a image in local and test out if we include the dev lib for fuzzy lib does the requirement satisfies, stay tuned for more updates.

Thanks @harshad16 for looking into it!

harshad16:
thanks for the patience, the pytest in the ci pipeline was fixed.

PR #2069, opened 2020-11-03
Raise an exception when the given record was not found

fridex:
... to respect API we have in other methods.

- [x] No


KPostOffice (approved):
LGTM :+1:

PR #2085, opened 2020-11-06
Relax constraint on inspection run result number to allow DM sync

pacospace:

Fixes: https://github.com/thoth-station/storages/issues/2073

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

Relax constraint on inspection run result number to allow DM sync


fridex (approved):
Hm, I wonder how these will be filled once an inspection finishes as we will have multiple entries with the same inspection id differing with `inspection_result_numer`. Haven't check that.

pacospace:
> Hm, I wonder how these will be filled once an inspection finishes as we will have multiple entries with the same inspection id differing with `inspection_result_numer`. Haven't check that.

As discussed in the chat, I closed this PR as it is no more required. Thanks!

Issue #1885, opened 2020-07-08
Introduce metrics about corrupted data [NOKR][tracker]


pacospace:
As Thoth Data Engineer,
I want to discover if any data is corrupted in the database, showing the analysis on Grafana, receive alarms from Prometheus and take action on it.

Acceptance Criteria:
- [x] ~~Create Dockerfile for postgresql10.0-rhel + amcheck that can be deployed.~~
- [x] Create query on database in thoth-storages (https://github.com/thoth-station/storages/issues/1586, https://github.com/thoth-station/storages/issues/1405)
- [x] Create metrics in metrics-exporter (https://github.com/thoth-station/metrics-exporter/issues/408)
- [x] Create alarm on Prometheus.

References:
- https://docs.google.com/document/d/1onhwI_3GGLfmAtZ74WDij_03kCk3Fn4ZoU0iDxCJkq4/edit

fridex:
https://github.com/thoth-station/metrics-exporter/issues/408

pacospace:
> [thoth-station/metrics-exporter#408](https://github.com/thoth-station/metrics-exporter/issues/408)

I linked that issue in this card [tracker] as third point already, as first we need an image with Postgres and that extension `amcheck`, probably RPM, to be created.

fridex:
> > [thoth-station/metrics-exporter#408](https://github.com/thoth-station/metrics-exporter/issues/408)
> 
> I linked that issue in this card [tracker] as third point already, as first we need an image with Postgres and that extension `amcheck`, probably RPM, to be created.

Not to maintain too many PG images, I think we could reuse the base image used in graph-backup-job.

CC @harshad16

KPostOffice:
I don't have access to `access.redhat.com` so I'm having trouble getting the yum repositories that I need in order to build amcheck from source.

pacospace:
@KPostOffice @goern

KPostOffice:
postgres10 actually has this extension available. No need to use a community build. `amcheck_next10` adds functionality of the amcheck available in postgres 11+ to postgres 10 and there is another build to backport the functionality to postgres9.  It should be as easy as enabling the module and running the command that I described in the doc.

pacospace:
This issue can be closed: https://gitlab.cee.redhat.com/data-hub/dh-monitoring/-/merge_requests/374

PR #2105, opened 2020-11-19
add function to run amcheck and check for db corruption

KPostOffice:

https://github.com/thoth-station/storages/issues/1885

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Runs amcheck to check for corruption

## Description

1. Check if amcheck is installed
    1. if amcheck is not installed, install it
1. run amcheck, if there is an error return true otherwise return false

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -5363,6 +5363,43 @@ class GraphDatabase(SQLBase):
         for method in self._CACHED_METHODS:
             method.cache_clear()
 
+    def is_database_corrupted(self) -> bool:
+        """
+        Run the amcheck extension to check for DB corruption, false negatives are possible but false positives are not.
+
+        amcheck documenation: https://www.postgresql.org/docs/10/amcheck.html
+
+        Returns
+        -------
+        bool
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -5363,6 +5363,43 @@ def cache_clear(self) -> None:
...
+
+        with self._session_scope() as session:
+            if session.execute(check_for_ext_query) == []:
+                session.execute(install_ext_query)
+
+            try:
+                session.execute(run_amcheck_query)
+                return False
+            except:
+                return True
```
fridex, line 5401: Just a minor comment. Can we log exception here? Might be useful for other failure scenarios.

fridex (commented):
:1st_place_medal: :1st_place_medal: :1st_place_medal: one minor comment, lgtm otherwise

thoth/storages/graph/postgres.py
KPostOffice: Yeah for sure ++

fridex (approved):
:100: :+1:

PR #2114, opened 2020-11-23
result must be cast to list

KPostOffice:
Noticed a bug that I didn't catch when testing because the extension was already added to my local DB


pacospace (approved):
lgtm

PR #2120, opened 2020-11-25
Queries needed for memory estimator

fridex:
These queries were needed for the memory estimator: https://github.com/fridex/graph-memory-estimator


fridex:
As it did not go in, let's keep it in the history.

Issue #2133, opened 2020-12-03
Some of the columns not present in kebechet_github_installations table


xtuchyna:
**Describe the bug**
When using `graph.get_kebechet_github_installations_count_per_is_active()` method, it seems that the `kebechet_github_installations.external_python_software_stack_id` column is missing.

This is also true according to the database inspection in `pgweb`, where I can see all columns from query present except these ones:

```
kebechet_github_installations.external_python_software_stack_id
kebechet_github_installations.external_software_environment_id
kebechet_github_installations.advised_python_software_stack_id
```

**To Reproduce**
1. set up the databse with the `pg_dump-20-12-01-00-02-19` dump
2. uses `thoth.storages` as follow:
```
>>> from thoth.storages import GraphDatabase
>>> graph = GraphDatabase()
>>> graph.connect()
The database schema is not in sync with library revisions, the current library revision heads: {'e7ebfc165a96'}, database heads: {'342d45db2dc2'}
>>> graph.get_kebechet_github_installations_count_per_is_active()
```

**Expected behavior**
A valid and correct query from the `graph.get_kebechet_github_installations_count_per_is_active()` with the return value of active kebechet github installations count.

**LOG**
```
>>> graph.get_kebechet_github_installations_count_per_is_active()
Traceback (most recent call last):
  File "/home/xtuchyna/.local/share/virtualenvs/storages-L98xb5vG/lib64/python3.8/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context
    self.dialect.do_execute(
  File "/home/xtuchyna/.local/share/virtualenvs/storages-L98xb5vG/lib64/python3.8/site-packages/sqlalchemy/engine/default.py", line 593, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.UndefinedColumn: column kebechet_github_installations.external_python_software_stack_id does not exist
LINE 2: ...st_run AS kebechet_github_installations_last_run, kebechet_g...
                                                             ^


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/xtuchyna/git/forks/storages/thoth/storages/graph/postgres.py", line 3482, in get_kebechet_github_installations_count_per_is_active
    session.query(KebechetGithubAppInstallations)
  File "/home/xtuchyna/.local/share/virtualenvs/storages-L98xb5vG/lib64/python3.8/site-packages/sqlalchemy/orm/query.py", line 3803, in count
[...]

PR #2134, opened 2020-12-03
Add get active installations method

xtuchyna:
Related to both https://github.com/thoth-station/mi-scheduler/pull/58
and https://github.com/thoth-station/common/pull/1054#discussion_r534069107

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

New `def get_kebechet_github_installations_active(self) -> List[str]:` method

## Description
The method returns the string list of all the repositories that have active kebechet installations

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3476,7 +3476,14 @@ class GraphDatabase(SQLBase):
             return False
 
     def get_active_kebechet_github_installations_repos(self) -> Dict[str, List[str]]:
-        """Get all active repositories names with active Kebechet installation."""
+        """Get all active repositories names with active Kebechet installation.
+        
+         Examples:
+        >>> from thoth.storages import GraphDatabase
+        >>> graph = GraphDatabase()
+        >>> graph.get_active_kebechet_github_installations_repos()
+        ['repository_foo_fullname', 'repository_bar_fullname', ...]
+        """
         with self._session_scope() as session:
             active_installations = (
                 session.query(KebechetGithubAppInstallations)
@@ -3484,7 +3491,7 @@ class GraphDatabase(SQLBase):
                 .all()
             )
 
-            return { i.slug:self.get_kebechet_github_installations_active_managers(slug=i.slug) for i in active_installations }
+            return [ i.slug for i in active_installations ]
 
     def get_kebechet_github_installations_count_per_is_active(self) -> int:
         """Return the count of active repos with Kebechet installation."""
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3475,12 +3475,13 @@ class GraphDatabase(SQLBase):
                 return True
             return False
 
-    def get_kebechet_github_installations_active(self) -> List[str]:
-        """Return all of the repositories wiht the active Kebechet installation."""
+    def get_kebechet_github_installations_per_is_active(self) -> List[str]:
+        """Get all active repositories names with active Kebechet installation."""
         with self._session_scope() as session:
             active_managers = (
                 session.query(KebechetGithubAppInstallations)
                 .filter(KebechetGithubAppInstallations.is_active.is_(True))
+                .all()
             )
             return [m.repo_name for m in active_managers]
 
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3475,6 +3475,15 @@ class GraphDatabase(SQLBase):
                 return True
             return False
 
+    def get_kebechet_github_installations_active(self) -> List[str]:
+        """Return all of the repositories wiht the active Kebechet installation."""
+        with self._session_scope() as session:
+            active_managers = (
+                session.query(KebechetGithubAppInstallations)
+                .filter(KebechetGithubAppInstallations.is_active.is_(True))
+            )
+            return [m.repo_name for m in active_managers]
+
     def get_kebechet_github_installations_count_per_is_active(self) -> int:
         """Return the count of active repos with Kebechet installation."""
         with self._session_scope() as session:
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3475,15 +3475,16 @@ class GraphDatabase(SQLBase):
                 return True
             return False
 
-    def get_kebechet_github_installations_per_is_active(self) -> List[str]:
+    def get_kebechet_github_installations_per_is_active(self) -> Dict[str, List[str]]:
         """Get all active repositories names with active Kebechet installation."""
         with self._session_scope() as session:
-            active_managers = (
+            active_installations = (
                 session.query(KebechetGithubAppInstallations)
                 .filter(KebechetGithubAppInstallations.is_active.is_(True))
                 .all()
             )
-            return [m.repo_name for m in active_managers]
+
+            return { i.slug:self.get_kebechet_github_installations_active_managers(slug=i.slug) for i in active_installations }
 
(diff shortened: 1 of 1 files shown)


xtuchyna:
NOTE: not yet tested, waiting for https://github.com/thoth-station/storages/issues/2133 to be solved

thoth/storages/graph/postgres.py
```diff
@@ -3475,6 +3475,15 @@ def update_kebechet_github_installations_on_is_active(self, slug: str) -> bool:
                 return True
             return False
 
+    def get_kebechet_github_installations_active(self) -> List[str]:
```
pacospace, line 3478: please read naming convention for queries: https://github.com/thoth-station/storages#query-naming-conventions-in-thoth

thoth/storages/graph/postgres.py
```diff
@@ -3475,6 +3475,15 @@ def update_kebechet_github_installations_on_is_active(self, slug: str) -> bool:
...
+        """Return all of the repositories wiht the active Kebechet installation."""
```
pacospace, line 3479: `Get all active repositories names with Kebechet installation.`

thoth/storages/graph/postgres.py
```diff
@@ -3475,6 +3475,15 @@ def update_kebechet_github_installations_on_is_active(self, slug: str) -> bool:
...
+        with self._session_scope() as session:
+            active_managers = (
+                session.query(KebechetGithubAppInstallations)
+                .filter(KebechetGithubAppInstallations.is_active.is_(True))
+            )
```
pacospace, line 3484: I think `.all()` is missing in query

thoth/storages/graph/postgres.py
xtuchyna: thx, didn't notice that!

thoth/storages/graph/postgres.py
xtuchyna: done, i've also added `active` to the docstring, is that alright?

thoth/storages/graph/postgres.py
```diff
@@ -3475,12 +3475,13 @@ def update_kebechet_github_installations_on_is_active(self, slug: str) -> bool:
                 return True
             return False
 
-    def get_kebechet_github_installations_active(self) -> List[str]:
-        """Return all of the repositories wiht the active Kebechet installation."""
+    def get_kebechet_github_installations_per_is_active(self) -> List[str]:
```
xtuchyna, line 3478: also @pacospace, what should be the return type? According to documentation it is `Dict[str, List[Any]]`,
should it be dict of repo_names with enabled managers then?

pacospace: your filter in this case, is `is_active` therefore you should have something like 

{'active': ['repo/1', 'repo/2'], inactive: [] }

but that is not what you want I think:

You want to have all repos that have active Kebechet, therefore I see two options 

1. => `get_kebechet_github_installations_all()` (Type one naming)


```
def get_kebechet_github_installations_per_is_active(self, is_active: Optional[bool] = None) -> List[str]:
"""Get all repositories names with Kebechet installation."""
            query = session.query(KebechetGithubAppInstallations)

              if is_active:
                     query = query.filter(KebechetGithubAppInstallations.is_active.is_(is_active))

             result = query.all()

         return [m.repo_name for m in result]
```

2. => `get_active_kebechet_github_installations_repos()` (Type seven naming)

```
def get_active_kebechet_github_installations_repos(self) -> List[str]:
"""Get all active repositories names with active Kebechet installation."""
            active_managers = (
                session.query(KebechetGithubAppInstallations)
                .filter(KebechetGithubAppInstallations.is_active.is_(is_active))
                .all()
            )
         return [m.repo_name for m in active_managers]
```


naming convention for `get_kebechet_github_installations_count_per_is_active` is not respected actually. Whatever is returning type number should be should have _count at the end => `get_active_kebechet_github_installations_repos_count_all` @saisankargochhayat (Type six naming)

in that case we can also generalize it actually:

```
 def get_kebechet_github_installations_repos_count_all(self,  is_active: Optional[bool] = None) -> int:
"""Get the number of repos with Kebechet installation."""
```

xtuchyna: Okay, changed accordingly to the type 7. Also I've changed the return value to be of { <repo>: [<managers>] }
plz have a l00k

thoth/storages/graph/postgres.py
```diff
@@ -3475,6 +3475,17 @@ def update_kebechet_github_installations_on_is_active(self, slug: str) -> bool:
...
+    def get_active_kebechet_github_installations_repos(self) -> Dict[str, List[str]]:
+        """Get all active repositories names with active Kebechet installation."""
+        with self._session_scope() as session:
+            active_installations = (
+                session.query(KebechetGithubAppInstallations)
+                .filter(KebechetGithubAppInstallations.is_active.is_(True))
+                .all()
+            )
+
+            return { i.slug:self.get_kebechet_github_installations_active_managers(slug=i.slug) for i in active_installations }
```
pacospace, line 3487: ?? we don't use one query inside another one the session needs to be one.

thoth/storages/graph/postgres.py
```diff
@@ -3475,6 +3475,17 @@ def update_kebechet_github_installations_on_is_active(self, slug: str) -> bool:
                 return True
             return False
 
+    def get_active_kebechet_github_installations_repos(self) -> Dict[str, List[str]]:
+        """Get all active repositories names with active Kebechet installation."""
```
pacospace, line 3479: Please enter an example of the method and the returned output. See: https://github.com/thoth-station/storages/blob/196fdfa3ba8c8b0f08ffce0bca434e0d3cc6efe8/thoth/storages/graph/postgres.py#L4138 for example

thoth/storages/graph/postgres.py
xtuchyna: So I've changed the function to return only repository names as originally intended, which should be proper to naming conventions

I hope I got it right this time @pacospace   ?

pacospace: Perfect, thanks @xtuchyna, if you need some more specific filter, you can create a generalized one, that take repos active with a certain manager active (set to True in the table)

xtuchyna: The `mi` should be able to autodetect which managers are active and which not, however in future it can be reasonable to really inspect if the returned active managers are really active or not.

pacospace:
please rebase @xtuchyna

xtuchyna:
done @pacospace

pacospace (approved):

thoth/storages/graph/postgres.py
```diff
@@ -3475,6 +3475,24 @@ def update_kebechet_github_installations_on_is_active(self, slug: str) -> bool:
...
+        """Get all active repositories names with active Kebechet installation.
+        
+         Examples:
+        >>> from thoth.storages import GraphDatabase
+        >>> graph = GraphDatabase()
+        >>> graph.get_active_kebechet_github_installations_repos()
+        ['repository_foo_fullname', 'repository_bar_fullname', ...]
+        """
+        with self._session_scope() as session:
+            active_installations = (
```
pacospace, line 3488: @xtuchyna you can also use directly `with_entities` to receive directly the specific parameter.
```
active_installations_repos = (
     session.query(KebechetGithubAppInstallations)
     .filter(KebechetGithubAppInstallations.is_active.is_(True))
     .with_entities(KebechetGithubAppInstallations.slug)
     .all()
)

```

fridex: This is better to do on the database side to avoid unnecessary serialization, rather than in python code.

PR #2141, opened 2020-12-07
Use proper naming convention of type 6

xtuchyna:
See discussion here https://github.com/thoth-station/storages/pull/2134#discussion_r535906381

- [X] Yes
- [ ] No

All the applications that are using this method should reference it with its new name

- [ ] Yes
- [X] No

Rename `get_kebechet_github_installations_count_per_is_active` to `get_active_kebechet_github_installations_repos_count_all` - the use of proper naming convention

## Description
 @pacospace wrote
> naming convention for get_kebechet_github_installations_count_per_is_active is not respected actually. Whatever is returning type number should be should have _count at the end => get_active_kebechet_github_installations_repos_count_all (Type six naming)

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3476,7 +3476,15 @@ class GraphDatabase(SQLBase):
             return False
 
     def get_active_kebechet_github_installations_repos_count_all(self) -> int:
-        """Return the count of active repos with Kebechet installation."""
+        """Return the count of active repos with Kebechet installation.
+        
+        Example:
+        >>> from thoth.storages import GraphDatabase
+        >>> graph = GraphDatabase()
+        >>> graph.connect()
+        >>> graph.get_active_kebechet_github_installations_repos_count_all()
+        165
+        """
         with self._session_scope() as session:
             count = (
                 session.query(KebechetGithubAppInstallations)
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3475,7 +3475,7 @@ class GraphDatabase(SQLBase):
                 return True
             return False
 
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -3475,7 +3475,7 @@ def update_kebechet_github_installations_on_is_active(self, slug: str) -> bool:
                 return True
             return False
 
-    def get_kebechet_github_installations_count_per_is_active(self) -> int:
+    def get_active_kebechet_github_installations_repos_count_all(self) -> int:
```
pacospace, line 3478: Please add also example output here :)

pacospace (commented):
Thanks @xtuchyna !! Please check with @saisankargochhayat , which components are affected by these changes, because after new release we need to have those adjusted already :)

saisankargochhayat:
> Thanks @xtuchyna !! Please check with @saisankargochhayat , which components are affected by these changes, because after new release we need to have those adjusted already :)

This is the only one I can think of - https://github.com/thoth-station/metrics-exporter/blob/a4cdd15bd45896a51ae167c42f09487f00b0b45b/thoth/metrics_exporter/jobs/kebechet.py#L39

pacospace:
> > Thanks @xtuchyna !! Please check with @saisankargochhayat , which components are affected by these changes, because after new release we need to have those adjusted already :)
> 
> This is the only one I can think of - https://github.com/thoth-station/metrics-exporter/blob/a4cdd15bd45896a51ae167c42f09487f00b0b45b/thoth/metrics_exporter/jobs/kebechet.py#L39

Thanks @saisankargochhayat :) :rocket:

pacospace (approved):

Issue #2147, opened 2020-12-09
Failed to sync adviser result with document id 'adviser-4d61a1b4.request


pacospace:
**Describe the bug**
There are documents stored in adviser on Ceph which are not documents to be synced. Are they supposed to be there? Should we filter those out when trying to retrieve adviser documents (e.g. advise-reporter is failing right not)

**To Reproduce**
1. Retrieve document ids for adviser

**Expected behavior**

**Screenshots**
```
{"name": "thoth.storages.sync", "levelname": "ERROR", "module": "sync", "lineno": 87, "funcname": "sync_adviser_documents", "created": 1607538402.1369357, "asctime": "2020-12-09 18:26:42,136", "msecs": 136.93571090698242, "relative_created": 1114575.3300189972, "process": 1, "message": "Failed to sync adviser result with document id 'adviser-4d61a1b4.request'\nTraceback (most recent call last):\n  File \"/opt/app-root/lib64/python3.8/site-packages/thoth/storages/sync.py\", line 81, in sync_adviser_documents\n    graph.sync_adviser_result(document)\n  File \"/opt/app-root/lib64/python3.8/site-packages/thoth/storages/graph/postgres.py\", line 4964, in sync_adviser_result\n    adviser_document_id = AdvisersResultsStore.get_document_id(document)\n  File \"/opt/app-root/lib64/python3.8/site-packages/thoth/storages/result_base.py\", line 73, in get_document_id\n    document_id = document[\"metadata\"].get(\"document_id\")\nKeyError: 'metadata'"}
```

**Additional context**

PR #2155, opened 2021-01-05
Filter out stored requests from listing

fridex:

Fixes: #2147

- [x] No


KPostOffice (commented):
This creates a generator even when the generator yields nothing, I'm not sure how you want to handle the pytest for this.

fridex:
> This creates a generator even when the generator yields nothing, I'm not sure how you want to handle the pytest for this.

Adjusted, PTAL.

pacospace (approved):
Thanks!!! LGTM

PR #2193, opened 2021-01-29
Move Thoth s2i specific env vars to software environment models

fridex:

This information correspond more to software environment than to package-extract run info. It will also save one join in https://github.com/thoth-station/adviser/issues/1631

- [x] No


pacospace (approved):
lgtm!

Issue #2196, opened 2021-01-29
Missing solver-ubi-8-py38 in database Ecosystem Solver table


pacospace:
**Describe the bug**
Missing solver-ubi-8-py38 in database but there is Missing solver-rhel-8-py38 in database. When querying for data in the database for specific solver solver-ubi-8-py38, none are identified. This might affect queries for adviser to look for data for rhel-8 mapped to ubi-8

**To Reproduce**
1. Have a look at the database

**Expected behavior**
In Ecosystem solver table we should find `solver-ubi-8-py38`

**Screenshots**

**Additional context**

pacospace:
I think it does affect queries, as we map to ubi always if someone input rhel. cc @fridex

goern:
@pacospace is this still relevant, if not please close it

goern:
is `curl -X GET "https://management.moc.thoth-station.ninja/api/v1/solvers" -H  "accept: application/json"` a valid test for this? is this covered by an integration test?

pacospace:
> is `curl -X GET "https://management.moc.thoth-station.ninja/api/v1/solvers" -H "accept: application/json"` a valid test for this? is this covered by an integration test?

This is an inconsistency in data in the database, not from solver CM. If the map from rhel 8 to ubi 8 is used in every method, we can close this issue.

goern:
sounds like a query we can run against the database, so that we check for consistency?

pacospace:
> sounds like a query we can run against the database, so that we check for consistency?

We have already that actually, number of solver in CM vs number of solver in database and we also planned an alarm there. So we can close this issue.

fridex:
If I recall correctly, we had `rhel` in the database that should not happen as it should be mapped to `ubi`. If `rhel` entries will be kept in the database, they will not cause any issues on the user side (as there is no solver installed). The incosystency will arise in monitoring when metrics-exporter exposes info about them. This might be confusing as we do not have "`rhel`" solvers per say. So the solution to this issue:

* make sure we map solver names correctly - https://github.com/thoth-station/storages/issues/2199
* check if we have `rhel` entries in the database -- if so, we can remove these entries using the purge-job

@pacospace could you please verify we have only `ubi` solvers reported in deployments and no `rhel` solvers?

pacospace:
[...]

PR #2197, opened 2021-01-29
Add queries and fix bug discovered

pacospace:

Related-To: https://github.com/thoth-station/storages/issues/2196

```
solvers:
['solver-fedora-31-py37', 'solver-fedora-31-py38', 'solver-fedora-32-py37', 'solver-fedora-32-py38', 'solver-ubi-8-py36']
number:
5
```

The new queries will be used in metrics-exporter, not to look for solver from CM but from database. In this way we know which solvers are present in the database, because in this moment only two solvers are shown from CM but in reality we have data from more solvers already in the database. 

In this way we can see already if there is an inconsistency in the data stored.

Moreover when we show users the solvers we have we should return the number from database.

See SLI report of today:

Parameter | Current value | Change since last week | Measurement Unit
-- | -- | -- | --
Solved Learning Rate | 0 | -363 | packages/hour
Solved packages | 610615 | +17338 | packages
Number of Solvers | 2 | 0 |  
Security Learning Rate | 0 | 0 | packages/hour
SI analyzed packages | 156807 | 0 | packages

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -1781,7 +1781,13 @@ class GraphDatabase(SQLBase):
         return query
 
     def get_ecosystem_solver_all(self) -> int:
-        """Get all solvers."""
+        """Get all solvers.
+        Examples:
+        >>> from thoth.storages import GraphDatabase
+        >>> graph = GraphDatabase()
+        >>> graph.get_ecosystem_solver_all()
+        ['solver-fedora-31-py38', 'solver-fedora-32-py37', 'solver-fedora-32-py38', 'solver-ubi-8-py36']
+        """
         with self._session_scope() as session:
             query = self._construct_get_solver_query(
                 session,
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -1780,7 +1780,7 @@ class GraphDatabase(SQLBase):
 
         return query
 
-    def get_ecosystem_solver_all(self) -> int:
+    def get_ecosystem_solver_all(self) -> List[str]:
         """Get all solvers.
         Examples:
         >>> from thoth.storages import GraphDatabase
@@ -1802,8 +1802,7 @@ class GraphDatabase(SQLBase):
             query = self._construct_get_solver_query(
                 session,
             )
-            count = query.count()
-            return count
+            return query.count()
 
     def get_solver_documents_count_all(self) -> int:
         """Get number of solver documents synced into graph."""
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -1767,6 +1767,38 @@ class GraphDatabase(SQLBase):
 
             return query.count()
 
+    def _construct_get_solver_query(
+        self,
+        session: Session
+    ) -> Query:
+        """Construct query to retrieve solvers."""
+        query = (
+            session.query(EcosystemSolver)
+            .filter(EcosystemSolver.os_name != 'rhel')   # Thoth considers rhel == ubi
+            .distinct(EcosystemSolver.solver_name)
+        )
+
+        return query
+
+    def get_ecosystem_solver_all(self) -> int:
+        """Get all solvers."""
+        with self._session_scope() as session:
+            query = self._construct_get_solver_query(
+                session,
+            )
+            solvers = query.with_entities(EcosystemSolver.solver_name).all()
+
+            return [s[0] for s in solvers]
+
+    def get_ecosystem_solver_count_all(self) -> int:
+        """Get number of solvers."""
+        with self._session_scope() as session:
+            query = self._construct_get_solver_query(
+                session,
+            )
+            count = query.count()
+            return count
+
     def get_solver_documents_count_all(self) -> int:
         """Get number of solver documents synced into graph."""
         with self._session_scope() as session:
@@ -4880,8 +4912,10 @@ class GraphDatabase(SQLBase):
                 ecosystem="python",
                 solver_name=solver_name,
                 solver_version=solver_version,
-                os_name=os_name,
-                os_version=os_version,
+                os_name=map_os_name(
+                    os_name
+                ),
+                os_version=normalize_os_version(os_version),
                 python_version=python_version,
             )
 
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -1767,6 +1767,43 @@ def get_si_unanalyzed_python_package_versions_count_all(
...
             return query.count()
 
+    def _construct_get_solver_query(
+        self,
+        session: Session
+    ) -> Query:
+        """Construct query to retrieve solvers."""
+        query = (
+            session.query(EcosystemSolver)
+            .filter(EcosystemSolver.os_name != 'rhel')   # Thoth considers rhel == ubi
```
fridex, line 1777: Why is this needed? We should prevent syncing data if they introduce RHEL vs UBI incosystency. Data in the database should be normalized and consistent.

fridex (commented):
> The new queries will be used in metrics-exporter, not to look for solver from CM but from database. In this way we know which solvers are present in the database, because in this moment only two solvers are shown from CM but in reality we have data from more solvers already in the database.

Can we have metrics for CM as well as based on these queries? The fact data are in the database does not implicate they can be used in recommendations - e.g. if solver is not enabled in CM and users asks for a recommendation, ingesting a missing package will not work (even though data/solver results for some packages found in the stack are synced in the database).

Related: https://github.com/thoth-station/thoth-application/issues/808

thoth/storages/graph/postgres.py
pacospace: > Why is this needed? We should prevent syncing data if they introduce RHEL vs UBI incosystency. Data in the database should be normalized and consistent.

we didn't do it immediately, so we have both rhel and ubi, and we have methods to map_os_name when you use query actually ubi8-py38 is missing, hence issue opened

pacospace: > > Why is this needed? We should prevent syncing data if they introduce RHEL vs UBI incosystency. Data in the database should be normalized and consistent.
> 
> we didn't do it immediately, so we have both rhel and ubi, and we have methods to map_os_name when you use query actually ubi8-py38 is missing, hence issue opened.

fridex: I think keeping this inconsistency might be not a good idea long-term. Can we create a database migration to fix data consistency?

pacospace:
> > The new queries will be used in metrics-exporter, not to look for solver from CM but from database. In this way we know which solvers are present in the database, because in this moment only two solvers are shown from CM but in reality we have data from more solvers already in the database.
> 
> Can we have metrics for CM as well as based on these queries? The fact data are in the database does not implicate they can be used in recommendations - e.g. if solver is not enabled in CM and users asks for a recommendation, ingesting a missing package will not work (even though data/solver results for some packages found in the stack are synced in the database).
> 
> Related: [thoth-station/thoth-application#808](https://github.com/thoth-station/thoth-application/issues/808)

Yes we can keep both on metricx-exporter side and differentiate. 

But if we finished solving for a solver and we remove it from CM, recommendation can still work, as adviser will check the database table and not the CM. 
From adviser point of view, which is user point of view, the database is the source, maybe we didn't solve all packages for a solver but in principle we have some data, because if I submit for solver fedora adviser would start and not fail for missing solver as it checks database and we do not deactivate a solver if we remove it from CM, so it is not part of the query. WDYT? Or adviser checks CM?

thoth/storages/graph/postgres.py
pacospace: > I think keeping this inconsistency might be not a good idea long-term. Can we create a database migration to fix data consistency?

agree it is not good. But we need to remove data and resync them, so that there is no in ecosystem solver, as we do not have a rhel solver, but only a ubi solver and the bug we are solving now on the sync side need to be used for re syncing

fridex:
> But if we finished solving for a solver and we remove it from CM, recommendation can still work, as adviser will check the database table and not the CM.
> From adviser point of view, which is user point of view, the database is the source, maybe we didn't solve all packages for a solver but in principle we have some data, because if I submit for solver fedora adviser would start and not fail for missing solver as it checks database and we do not deactivate a solver if we remove it from CM, so it is not part of the query. WDYT? Or adviser checks CM?

Yes, the recommendation might work based on the current logic but it does not guarantee the data are up2date and that these recommendations can be used. We can still serve recommendations, but users should be aware that the data are not up to date.

E.g. let's say fedora:32 goes EOL so we turn off solver-fedora-32-py38 (by removing it from the CM) and a new `tensorflow==2.5.0` is released after that. A user requests to run `tensorflow==2.5.0` on fedora:32+py38, the current logic tells the user the package is not present and will be solved+analyzed - but that's not actually true as the solver is not installed (moreover it will produce errors on the missing package workflow which would need to be handled). We could produce some messages to users that the data are no longer up2date, but I would rather go with https://github.com/thoth-station/thoth-application/issues/808 where we correctly shut down recommendations for specific runtime environments - remove entries from the configmap and remove data from the database. This will enable us to ingest data for new runtime environment (e.g. fedora release +3) and keep the service "rolling" with Python/runtime environments. All this can go to release notes ("fedora XX is no longer supported", "fedora +3 is supported and data ingestion is turned on").

thoth/storages/graph/postgres.py
fridex: I think a simple UPDATE SQL query that would work on os_name = rhel/ubi should work fine. Not to remove data which might cause issues with already linked entries in the database.

pacospace: We do this directly on the database then, thanks @fridex, something more on this PR? :)

pacospace:
@fridex PTAL ☕

thoth/storages/graph/postgres.py
fridex: I think the rhel string should not be placed here. If we decide solvers `rhel==ubi` are equivalent, we should update the data in the database to keep them consistent following https://github.com/thoth-station/storages/issues/2199

pacospace: Adjusted in dc0267a8d1c9e33e5b3d6909ccda917bd6fb8487, Thanks!

fridex (approved):
lgtm, thanks! :+1:

PR #2217, opened 2021-02-03
Adjust query for user software stack

pacospace:

Screenshot from 2021-02-03 09-43-38

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Add query to correctly retrieve number of user software stacks. Recently we introduced ExternalPythonSoftwareStack table only for users, therefore the query currently used in metrics-exporter reports always 0.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3945,22 +3945,18 @@ class GraphDatabase(SQLBase):
 
         return software_stack
 
-    def get_python_software_stack_count_all(self, software_stack_type: str, distinct: bool = False) -> int:
+    def get_python_software_stack_count_all(self, is_external: bool = False, software_stack_type: Optional[str] = None, distinct: bool = False) -> int:
         """Get number of Python software stacks available filtered by type."""
         with self._session_scope() as session:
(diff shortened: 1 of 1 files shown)


fridex (commented):
In other places, we parametrize these queries using `is_external`:

https://github.com/thoth-station/storages/blob/78cde6ec8bfa9315b38584a9faaaad017f2caa14/thoth/storages/graph/postgres.py#L2650

Would it make sense to keep this API? I think it somehow violates the naming convention, but it might be good to keep it consistent.

pacospace:
> In other places, we parametrize these queries using `is_external`:
> 
> https://github.com/thoth-station/storages/blob/78cde6ec8bfa9315b38584a9faaaad017f2caa14/thoth/storages/graph/postgres.py#L2648-L2652
> 
> Would it make sense to keep this API? I think it somehow violates the naming convention, but it might be good to keep it consistent.

Right, thanks! Let me squash the query with the other one :)

fridex (approved):
lgtm :100: :+1:

pacospace:
Based on https://github.com/thoth-station/thoth-application/issues/747, merge it manually this time

PR #2220, opened 2021-02-03
add marker file PEP561

KPostOffice:

…

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

https://www.python.org/dev/peps/pep-0561/

## Description

It looks like this was meant to be included based on, https://github.com/thoth-station/storages/blob/8448e20488d3e3229b2d553cedd9e803e7149534/setup.py#L88


fridex (approved):
Thanks! :+1:

fridex:
@KPostOffice A change in setup.py will be also needed.

KPostOffice:
> @KPostOffice A change in setup.py will be also needed.

https://github.com/thoth-station/storages/blob/8448e20488d3e3229b2d553cedd9e803e7149534/setup.py#L88

I think this is all that is required. It is already present in `setup.py`.

fridex:
> > @KPostOffice A change in setup.py will be also needed.
> 
> https://github.com/thoth-station/storages/blob/8448e20488d3e3229b2d553cedd9e803e7149534/setup.py#L88
> 
> @fridex I think this is all that is required. It is already present in `setup.py`.

I wonder how is it possible we missed it in the package tree then. Thanks!

PR #2228, opened 2021-02-25
Introduce a query for obtaining solver document ids based on package

fridex:

With this query, we are able to obtain all the solver document ids that correspond to solving the given package.

- [x] No

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -799,6 +799,46 @@ class GraphDatabase(SQLBase):
         kwargs.pop("self", None)  # static method
         return self.__class__._construct_python_packages_query(**kwargs)
 
+    def get_solver_document_id_for_package_version(
+        self,
+        package_name: str,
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -799,6 +799,46 @@ def _construct_solved_python_packages_query(
         kwargs.pop("self", None)  # static method
         return self.__class__._construct_python_packages_query(**kwargs)
 
+    def get_solver_document_id_for_package_version(
```
pacospace, line 802: Shouldn't be just another filter for: https://github.com/thoth-station/storages/blob/b29d1606d5dce208e18012e456900810293284ed/thoth/storages/graph/postgres.py#L4323

fridex: +1, good catch. Will move them there.

fridex:
/close

I think we do not need to ship this functionality.

PR #2231, opened 2021-03-02
Add query for number of rows in alembic table

pacospace:

Related-To: https://github.com/thoth-station/metrics-exporter/issues/573

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Introduce query for number of rows in alembic version table

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -383,6 +383,14 @@ class GraphDatabase(SQLBase):
 
         return is_up2date
 
+    def get_alembic_version_count_all(self) -> int:
+        """Get number of records in alembic version table (1 expected)."""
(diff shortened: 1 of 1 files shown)


KPostOffice (commented):
Can we query a single column when we count, like `id`? I'm not sure the exact schema of the `alembic_version` table.

pacospace:
> Can we query a single column when we count, like `id`? I'm not sure the exact schema of the `alembic_version` table.

alembic_version table has only one column called `version_num`, is that ok @KPostOffice that table is not created by us directly

pacospace:
@KPostOffice @fridex can we get this in?

fridex (approved):
lgtm :+1:

PR #2244, opened 2021-03-12
Do not enforce Python version being set for non-external images

fridex:

- [x] No

Related-to: https://github.com/thoth-station/graph-sync-job/issues/562


harshad16 (commented):
/lgtm
thanks for fixing this.

PR #2254, opened 2021-03-16
change err to warning

KPostOffice:
Change error to warning so sentry issues are not created

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -5797,7 +5797,7 @@ class GraphDatabase(SQLBase):
                 session.execute(run_amcheck_query)
                 return False
             except Exception as e:
-                _LOGGER.error(e)
+                _LOGGER.warning(e)
                 return True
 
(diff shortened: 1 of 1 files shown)


goern:
hmm, we require amcheck dont we? So I would leave it as an error and eliminate the root cause of the issue we have seen.

KPostOffice:
This is separate from `amcheck` missing. This command errors when the query fails.

fridex:
> This is separate from `amcheck` missing. This command errors when the query fails.

Is this related to reporting insufficient privileges to run amcheck?

KPostOffice:
@fridex Yeah

harshad16:
This PR can be closed.
The insufficient privilege issue was resolved.
Please review and comment on: https://github.com/thoth-station/storages/issues/2250#issuecomment-801633584

fridex:
> This PR can be closed.

OK, let's close this. This suppresses alerting, if the issue was fixed, the alert will not be fired.

PR #2258, opened 2021-03-19
Fix/avoid duplicate heads

pacospace:

Screenshot from 2021-03-19 16-58-59

The old method sometimes expose two heads, which can trigger alarms that are not required.
This PR modify the method to retrieve database head, not using alembic version methods but directly querying the table in the datbaase.


fridex (approved):
lgtm :+1:

PR #2263, opened 2021-03-19
Create index for depends_on considering is_missing and extras

fridex:

- [x] No

This query is used in resolver and the current implementation requires a sequantial scan which is not well performing.


KPostOffice (approved):
LGTM :+1:

pacospace (approved):
lgtm!

PR #2281, opened 2021-04-12
fix iter dates method

pacospace:

### DIFF
--- a/thoth/storages/result_base.py
+++ b/thoth/storages/result_base.py
@@ -94,6 +94,8 @@ class ResultStorageBase(StorageBase):
             end_date = date.today() + timedelta(days=1)  # Today inclusively.
         elif end_date < start_date:
             raise ValueError("end_date cannot precede start_date")
+        elif end_date == start_date:
+            end_date = start_date + timedelta(days=1)
 
         walker = start_date
         step = timedelta(days=1)


fridex (approved):
nice catch 👍🏻

fridex:
Wait... What is the intention here? The original code looks correct to me - if there is no `end_date` date supplied, today+1 is considered as the `end_date`.

pacospace:
> Wait... What is the intention here? The original code looks correct to me - if there is no `end_date` date supplied, today+1 is considered as the `end_date`.

How do I take the data of the same date? for example 2021-february-26. Either we modify that line or there is a bug when `start_date == end_date`, no data are collected, unless we modify `while walker <= end_date`

fridex:
The following code should give back all the results starting year 2021 to this date:

```python
adapter.iterate_results(start_date=date(year=2021, month=1, day=1), end_date=None)
```

With the patch applied it would get back only results for this equivalent call:

```python
adapter.iterate_results(start_date=date(year=2021, month=1, day=1), end_date=date(year=2021, month=1, day=2))
```

pacospace:
> The following code should give back all the results starting year 2021 to this date:
> 
> ```python
> adapter.iterate_results(start_date=date(year=2021, month=1, day=1), end_date=None)
> ```
> 
> With the patch applied it would get back only results for this equivalent call:
> 
> ```python
> adapter.iterate_results(start_date=date(year=2021, month=1, day=1), end_date=date(year=2021, month=1, day=2))
> ```
what about `start_date == end_date`?

fridex:
> > Wait... What is the intention here? The original code looks correct to me - if there is no `end_date` date supplied, today+1 is considered as the `end_date`.
> 
> How do I take the data of the same date? for example 2021-february-26:

The `end_date` is not inclusive:

```python
adapter.iterate_results(start_date=date(year=2021, month=2, day=26), end_date=date(year=2021, month=2, day=27))
```

Reads as "get results starting 2021-02-26 till 2021-02-27".

pacospace:
> > > Wait... What is the intention here? The original code looks correct to me - if there is no `end_date` date supplied, today+1 is considered as the `end_date`.
> > 
> > 
> > How do I take the data of the same date? for example 2021-february-26:
> 
> The `end_date` is not inclusive:
> 
> ```python
> adapter.iterate_results(start_date=date(year=2021, month=2, day=26), end_date=date(year=2021, month=2, day=27))
> ```
> 
> Reads as "get results starting 2021-02-26 till 2021-02-27".

So it is expected to give no results if start_date == end_date? I would add a condition then when they are equal to fall back to the case you just described, wdyt?

pacospace:
> > > > Wait... What is the intention here? The original code looks correct to me - if there is no `end_date` date supplied, today+1 is considered as the `end_date`.
> > > 
> > > 
> > > How do I take the data of the same date? for example 2021-february-26:
> > 
> > 
> > The `end_date` is not inclusive:
> > ```python
> > adapter.iterate_results(start_date=date(year=2021, month=2, day=26), end_date=date(year=2021, month=2, day=27))
> > ```
> > 
> > 
> > Reads as "get results starting 2021-02-26 till 2021-02-27".
> 
> So it is expected to give no results if `start_date == end_date`? I would add a condition then when they are equal to fall back to the case you just described, wdyt?

added 
```
        elif end_date == start_date:
            end_date = start_date + timedelta(days=1)
```

fridex:
> > So it is expected to give no results if `start_date == end_date`? I would add a condition then when they are equal to fall back to the case you just described, wdyt?
> 
> added
> 
> ```
>         elif end_date == start_date:
>             end_date = start_date + timedelta(days=1)
> ```

I would expect it to behave differently to be consistent. In Python, you get the following range in the standard library:

```pytohn
>>> list(range(0, 5))
[0, 1, 2, 3, 4]
```

pacospace:
> > > So it is expected to give no results if `start_date == end_date`? I would add a condition then when they are equal to fall back to the case you just described, wdyt?
> > 
> > 
> > added
> > ```
> >         elif end_date == start_date:
> >             end_date = start_date + timedelta(days=1)
> > ```
> 
> I would expect it to behave differently to be consistent. In Python, you get the following range in the standard library:
> 
> ```
> >>> list(range(0, 5))
> [0, 1, 2, 3, 4]
> ```

Totally make sense!! Thanks 🍻

PR #2292, opened 2021-04-30
Provide methods for deleting data from the database

fridex:

Related: https://github.com/thoth-station/thoth-application/issues/808

- [x] No

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -5941,3 +5941,74 @@ class GraphDatabase(SQLBase):
         bloat_data = [table for table in result if table["pct_bloat"] is not None or table["mb_bloat"] is not None]
 
         return bloat_data
+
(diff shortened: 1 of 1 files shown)


fridex:
@pacospace PTAL I tested the queries locally, but extra pair of eyes when deleting data are always valuable 🔢  👓 👀

pacospace (commented):
lgtm! 💯 one question regarding deletion, for example for solver, are the data deleting also relationships? like package version and pacakge version for a solver?

fridex:
> lgtm!  one question regarding deletion, for example for solver, are the data deleting also relationships? like package version and pacakge version for a solver?

Yes, they should be - we have `ondelete="CASCADE"` all over the place (tried it locally). But worth to make a dump before trying these operations in a deployment.

pacospace (approved):
Thanks 💯 👀 did not spot anything, but it is also friday 🍻

PR #2293, opened 2021-04-30
get or create github kebechet installation

KPostOffice:
Function wasn't working if the entry didn't exist in the database. Here we create if it doesn't exist yet.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3711,22 +3711,25 @@ class GraphDatabase(SQLBase):
     def update_kebechet_installation_using_files(
         self,
         slug: str,
+        installation_id: str,
         requirements: Optional[dict] = None,
         requirements_lock: Optional[dict] = None,
         thoth_config: Optional[dict] = None,
+        private: bool = False,
     ):
         """Update info about kebechet installation."""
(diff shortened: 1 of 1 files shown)


pacospace:
Related-To: https://github.com/thoth-station/user-api/issues/1181

thoth/storages/graph/postgres.py
```diff
@@ -3711,48 +3711,80 @@ def get_kebechet_github_installations_info_for_python_package_version(
...
-                external_software_environment=software_env,
-                info_manager="info" in all_managers,
-                pipfile_requirements_manager="pipfile-requirements" in all_managers,
-                update_manager="update" in all_managers,
-                version_manager="version" in all_managers,
-                thoth_advise_manager="thoth-advise" in all_managers,
-                thoth_provenance_manager="thoth-provenance" in all_managers,
+            item = (
+                session.query(KebechetGithubAppInstallations)
+                .filter(KebechetGithubAppInstallations.installation_id == installation_id)
```
pacospace, line 3749: `slug` is unique in the database and not `installation_id`, because user can install, uninstall the app and a new installation_id would be created in that case (I assume)

pacospace (approved):
Thanks! 💯

PR #2302, opened 2021-05-04
Make at least one filter mandatory when deleting package-extract runs

fridex:

Related: …

- [x] No


pacospace (approved):
lgtm!

PR #2314, opened 2021-06-01
Fix only_if_package_seen logic handling

fridex:

- [x] No

- [x] Yes

## Description

Wrong indentation causes wrong handling of `only_if_package_seen` logic. The given package is never created if `only_if_package_seen=False`.


harshad16 (approved):
lgtm, thanks for the fix :+1:

PR #2345, opened 2021-06-11
Sync solver name, no mapping required in solver sync logic

pacospace:

Related-To: https://github.com/thoth-station/storages/issues/2196
Sync solver name, no mapping required in solver sync logic

we don't have rhel solver names
we don't need to map from ubi to rhel if solver name has ubi

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -5091,7 +5091,7 @@ class GraphDatabase(SQLBase):
                 ecosystem="python",
                 solver_name=solver_name,
                 solver_version=solver_version,
-                os_name=map_os_name(os_name),
+                os_name=os_name,
                 os_version=normalize_os_version(os_name, os_version),
                 python_version=python_version,
             )


thoth/storages/graph/postgres.py
```diff
@@ -5091,7 +5091,7 @@ def sync_solver_result(self, document: dict, *, force: bool = False) -> None:
                 ecosystem="python",
                 solver_name=solver_name,
                 solver_version=solver_version,
-                os_name=map_os_name(os_name),
+                os_name=os_name,
```
fridex, line 5094: I'm not sure what issue will this address. The `map_os_name` method maps `ubi`->`rhel` in solvers:

https://github.com/thoth-station/common/blob/5b96cd3c2ddf4856426f69f87421215643f3f4cc/thoth/common/helpers.py#L182-L187

As we do not have "solver-ubi-*" names ([quay](https://quay.io/organization/thoth-station)), this operation is (correctly) a noop in the syncing logic. The call removed just makes sure that if a solver named  `solver-ubi-*` is accidentally registered in the system, the database entries will hold correctly remapped entries as `solver-rhel-*` would be used.

pacospace: > I'm not sure what issue will this address. The `map_os_name` method maps `ubi`->`rhel` in solvers:
> 
> https://github.com/thoth-station/common/blob/5b96cd3c2ddf4856426f69f87421215643f3f4cc/thoth/common/helpers.py#L182-L187
> 
> As we do not have "solver-ubi-*" names ([quay](https://quay.io/organization/thoth-station)), this operation is (correctly) a noop in the syncing logic. The call removed just makes sure that if a solver named `solver-ubi-*` is accidentally registered in the system, the database entries will hold correctly remapped entries as `solver-rhel-*` would be used.

So we need the opposite mapping: https://github.com/thoth-station/storages/issues/2196#issuecomment-859491353
as we want to have `ubi` in the database and not `rhel`

pacospace: We can close this PR, as the logic is correctly mapping ubi entries to rhel.

Issue #2374, opened 2021-06-29
Make platform constant in sources


fridex:
**Is your feature request related to a problem? Please describe.**

With recent issues discovered on the database side (thanks to @harshad16), we could make platform constant - as we do not plan to support other platforms in near future, having just linux-86_64 should work fine. It will reduce pressure on the database queries and will save few gigs of disk space given the size of tables (`depends_on`).

**Describe the solution you'd like**

Do not store platform in `depends_on` table, but rather make it constant (`linux-x86_64`) in sources.

pacospace:
@harshad16 already solved this issue, I think it can be closed!

pacospace:
Let's close it, feel to reopen :)

PR #2375, opened 2021-06-29
Deprecated the platform column form the dependsOn table

harshad16:
Deprecated the platform column form the dependsOn table

Related-to: https://github.com/thoth-station/storages/issues/2374
https://github.com/thoth-station/thoth-application/issues/1766

- [x] Yes

- [x] Yes

## Description

This would remove the use of platform column form dependsOn table.


thoth/storages/graph/postgres.py
```diff
@@ -2317,17 +2316,11 @@ def get_python_package_version_dependents_all(
                         "marker_evaluation_result": entry[4],
                         "marker": entry[5],
                         "extra": entry[6],
-                        "platform": entry[7],
```
pacospace, line 2320: Not sure which components use this query actually

thoth/storages/graph/postgres.py
```diff
@@ -5231,11 +5219,6 @@ def sync_solver_result(self, document: dict, *, force: bool = False) -> None:
         os_name = solver_info["os_name"]
         os_version = solver_info["os_version"]
         python_version = solver_info["python_version"]
-        # Older solver documents did not provide platform explictly.
-        platform = document["result"].get("platform") or "linux-x86_64"
-
-        if not self.python_package_version_depends_on_platform_exists(platform=platform):
```
pacospace, line 5237: I think we want to keep this check in any case, because when we will have more platforms we need to take care of that column. wdyt?

thoth/storages/graph/postgres.py
harshad16: revsolver uses this query
https://github.com/thoth-station/revsolver/blob/e928e373428c909638339198c7369a134616e851/app.py#L64

thoth/storages/graph/postgres.py
harshad16: oh you mean like lets leave the check, and decide in future if we wanna add that to the column.
understood. will get these back.

thoth/storages/graph/postgres.py
pacospace: thanks! But the platform itself is not required, so removing it, won't break it in theory. ack?

thoth/storages/graph/postgres.py
pacospace: Yes exactly, even if we don't have other platforms for now, better to keep it!

thoth/storages/graph/postgres.py
harshad16: yes, correct it wont, break it.

pacospace (approved):
lgtm!!! 💯

PR #2378, opened 2021-06-29
Remove qeb-hwt integration

pacospace:

Related-To: https://github.com/thoth-station/user-api/issues/1454

### DIFF
--- a/thoth/storages/graph/enums.py
+++ b/thoth/storages/graph/enums.py
@@ -81,7 +81,6 @@ class ThothAdviserIntegrationEnum(Enum):
     CLI = auto()
     KEBECHET = auto()
     S2I = auto()
-    GITHUB_APP = auto()
     JUPYTER_NOTEBOOK = auto()
 
 
--- a/thoth/storages/graph/enums.py
+++ b/thoth/storages/graph/enums.py
@@ -75,15 +75,6 @@ class QuerySortTypeEnum(Enum):
     PACKAGE_VERSION = "package_version"
 
 
-class ThothAdviserIntegrationEnum(Enum):
-    """Class for source type enum."""
-
-    CLI = auto()
-    KEBECHET = auto()
-    S2I = auto()
-    JUPYTER_NOTEBOOK = auto()
-
-
 class PlatformEnum(Enum):
     """Class for platform enum."""
 
--- a/thoth/storages/graph/models.py
+++ b/thoth/storages/graph/models.py
@@ -31,6 +31,8 @@ from sqlalchemy.dialects.postgresql import ENUM
 from sqlalchemy import UniqueConstraint
 from sqlalchemy import Index
 
+from thoth.common.enums import ThothAdviserIntegrationEnum
+
 from .models_base import BaseExtension
 from .models_base import Base
 from .models_base import get_python_package_version_index_combinations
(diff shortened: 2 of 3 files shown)


KPostOffice:
Hmm. I worry about how this will change the values of the surrounding enums if we store them anywhere. Can we explicitly set them here so that they don't change?

pacospace:
> Hmm. I worry about how this will change the values of the surrounding enums if we store them anywhere. Can we explicitly set them here so that they don't change?

Sorry I didn't get what needs to be done :) you mean removing auto()?

fridex:
Calls to `auto` make [incremental values for enums](https://docs.python.org/3/library/enum.html#using-automatic-values):

```python
    CLI = auto()                                     # ==1
    KEBECHET = auto()                                # ==2
    S2I = auto()                                     # ==3
    GITHUB_APP = auto()                              # ==4
    JUPYTER_NOTEBOOK = auto()                        # ==5
```

by removing `GITHUB_APP`, `JUPYTER_NOTEBOOK` will be equal to 4. That means any stored values of the enum affected will change semantics.

pacospace:
> Calls to `auto` make [incremental values for enums](https://docs.python.org/3/library/enum.html#using-automatic-values):
> 
> ```python
>     CLI = auto()                                     # ==1
>     KEBECHET = auto()                                # ==2
>     S2I = auto()                                     # ==3
>     GITHUB_APP = auto()                              # ==4
>     JUPYTER_NOTEBOOK = auto()                        # ==5
> ```
> 
> by removing `GITHUB_APP`, `JUPYTER_NOTEBOOK` will be equal to 4. That means any stored values of the enum affected will change semantics.

Thanks @fridex!! so what should we do now? should we move to something like: https://github.com/thoth-station/storages/blob/59168a39eb5480e59d2f095e2e42f69933efd589/thoth/storages/graph/enums.py#L62.

Also why do we have two classes for integrations? One in storages and one in common? should we just keep one to avoid inconsistencies? https://github.com/thoth-station/common/blob/a88ee9b7ccfff937920d9ee4016c58ffbb4eb7a4/thoth/common/enums.py#L23

fridex:
> Also why do we have two classes for integrations? One in storages and one in common? should we just keep one to avoid inconsistencies? https://github.com/thoth-station/common/blob/a88ee9b7ccfff937920d9ee4016c58ffbb4eb7a4/thoth/common/enums.py#L23

That would be great 👍🏻

pacospace:
> > Also why do we have two classes for integrations? One in storages and one in common? should we just keep one to avoid inconsistencies? https://github.com/thoth-station/common/blob/a88ee9b7ccfff937920d9ee4016c58ffbb4eb7a4/thoth/common/enums.py#L23
> 
> That would be great 👍🏻

Then we could keep it in storages I guess as the content is mapped to database in general? wdyt?

fridex:
> > > Also why do we have two classes for integrations? One in storages and one in common? should we just keep one to avoid inconsistencies? https://github.com/thoth-station/common/blob/a88ee9b7ccfff937920d9ee4016c58ffbb4eb7a4/thoth/common/enums.py#L23
> > 
> > 
> > That would be great 👍🏻
> 
> Then we could keep it in storages I guess as the content is mapped to database in general? wdyt?

Given its use in thoth-common, I would say the enum is more suitable to stay in thoth-common and let thoth-storages import it from there.

KPostOffice (commented):
/lgtm does this need a new release of thoth-common or is it gtg?

PR #2384, opened 2021-07-01
add runtime_environment_name to KebechetGithubInstallation

KPostOffice:

Required for differentiating between overlay environments in database.

It feels kinda wrong to add multiple entries in this table for a single app installation, but I am uncertain of the benefit and feasibility of adding another table.


pacospace (commented):
lgtm!!! 💯

thoth/storages/graph/models.py
```diff
@@ -1585,12 +1585,18 @@ class KebechetGithubAppInstallations(Base, BaseExtension):
...
+        db.UniqueConstraint('slug', 'runtime_environment_name'),
+    )
+
     id = Column(Integer, primary_key=True, autoincrement=True)
     slug = Column(Text, nullable=False, unique=True)
     repo_name = Column(Text, nullable=False)
     private = Column(Boolean, nullable=False)
     installation_id = Column(Text, nullable=False)
     is_active = Column(Boolean, nullable=False)
+    runtime_environment_name = Column(Text, nullable=True)
```
pacospace, line 1599: alembic migration is missing.

KPostOffice: Hey, I added it but it seems like there is too much going on in the migration could you take a look?

KPostOffice:
@pacospace is this good to go?

thoth/storages/graph/postgres.py
```diff
@@ -3706,6 +3709,7 @@ def get_kebechet_github_installations_active_managers(self, slug: str) -> list:
                 session.query(KebechetGithubAppInstallations)
                 .filter(KebechetGithubAppInstallations.is_active.is_(True))
                 .filter(KebechetGithubAppInstallations.slug == slug)
+                .filter(KebechetGithubAppInstallations.runtime_environment_name == runtime_environment_name)
```
pacospace, line 3712: This should be optional if runtime_environemnt is optional.

thoth/storages/data/alembic/versions/5c1e7425ce60_added_entry_in_.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision = '65ae36e5c38a'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('kebechet_github_installations', sa.Column('runtime_environment_name', sa.Text(), nullable=True))
+    op.create_unique_constraint(None, 'kebechet_github_installations', ['slug', 'runtime_environment_name'])
+    op.create_unique_constraint(None, 'python_package_version_entity_rule', ['id'])
```
pacospace, line 23: This does not seem to be part of this migration right? Also in Downgrade

KPostOffice: Removed

thoth/storages/graph/postgres.py
KPostOffice: Runtime environment is a nullable column, so this is a valid query that should always return a single row. I can remove the default value to make this more clear but it will require changes to packages using this function.

thoth/storages/data/alembic/versions/5c1e7425ce60_added_entry_in_.py
pacospace: there is also `python_package_version_entity_rule` not from this migration

KPostOffice: @pacospace I forgot to push changes. It should be removed now.

pacospace (approved):
Lgtm! 💯

PR #2410, opened 2021-07-27
Add packages and populate with module names

tlegen-k:
https://github.com/thoth-station/datasets/pull/36 
…

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Add module names without namespaced calls from results to packages list to sync with database.


tlegen-k:
Hi @pacospace 
Would you be able to take a look?
Next step will be to create new table linking PythonPackageVersion Table with the new table.

thoth/storages/graph/postgres.py
```diff
@@ -5477,6 +5487,17 @@ def sync_solver_result(self, document: dict, *, force: bool = False) -> None:
                                 marker_evaluation_result=dependency.get("marker_evaluation_result", True),
                             )
 
+            # Sync imoprted package from result:tree except for test module, all the namespaced calls
+            # modules starting with __ also excluded
+            # For Thoth-solver-dataset-v2 ["result"]["tree"]["packages"] has only one module to import
+            import_package =[]
+            for package_call in document["result"]["tree"]["packages"]:
+                if package_call.count(".") == 0 && package_call[:4] != "test"
+                    && package_call[0] != "_":
```
fridex, line 5496:
```suggestion
                    and package_call[0] != "_":
```

fridex: Could you give examples for entries that are skipped?

fridex (commented):
Looking at this, I'm not sure if we do not break `get_or_create` logic and create redundant entries (one with packages and the other without).

thoth/storages/graph/postgres.py
tlegen-k: > Could you give examples for entries that are skipped?

Hi @fridex, thank you for helping!
Yes, when exploring [Thoth solver dataset v2 ](https://www.kaggle.com/thothstation/thoth-solver-dataset-v10?select=thoth-solver-dataset-v2.0) for report with documet_id = solver-rhel-8-py38-210713042150-50168537c03f93c5 tree:packages looks like:
```
['_plotly_future_',
 '_plotly_utils',
 'plotly',
 'plotlywidget',
 '_plotly_utils.colors',
 'plotly.colors',
 'plotly.data',
 'plotly.express',
 'plotly.figure_factory',
 'plotly.graph_objs',
 'plotly.io',
 'plotly.matplotlylib',
 'plotly.offline',
 'plotly.plotly',
 'plotly.validators',
 'plotly.graph_objs.area',
...
]
```

Here we would like to skip all hidden entry imports that start with '_'.
Also AFAIK it is the only one that has two modules, but `plotlywidget` is not on pypi.org.

thoth/storages/graph/postgres.py
```diff
@@ -4116,6 +4118,7 @@ def _create_python_package_version(
         python_version: Union[str, None],
         python_package_metadata_id: int = None,
         sync_only_entity: bool = False,
+        import_package: str,
```
pacospace, line 4121: mandatory parameter cannot be after optional one :) is it optional? then I would put `Optional[str] = None`

thoth/storages/graph/models.py
```diff
@@ -72,7 +72,7 @@ class PythonPackageVersion(Base, BaseExtension):
     )
     is_missing = Column(Boolean, nullable=False, default=False)
     provides_source_distro = Column(Boolean, nullable=False, default=True)
-
+    import_package = Column(Text, nullable=False)
```
pacospace, line 75: nullable should be True, initially the database won't have these entries and we cannot assign a default in this case, wdyt?

tlegen-k: > nullable should be True, initially the database won't have these entries and we cannot assign a default in this case, wdyt?

I agree, corrected in this commit d1a6c6a850b5e8f3e2577648f8c348225f492f8c.
Thanks for pointing!

thoth/storages/graph/postgres.py
tlegen-k: thank you)
I think it should be mandatory in this context 
corrected with c7996586173df63d59a6d46ac5a5063bc89bca20

tlegen-k:
Hi @pacospace and @fridex,

Regards possible number of modules in tree:packages, apparently there are more cases when there is more than 1 module.
As an example this [notebook](https://colab.research.google.com/drive/1jF84zOzVTd5ThW5kvSE7C8fz4M9xYuwf?usp=sharing).

To give part of nonunique imports:
```
[['plotly', 'plotlywidget'],
 ['chart_studio', 'plotly', 'plotlywidget'],
 ['Cython', 'pyximport'],
 ['pkg_resources', 'setuptools'],
 ['plotly', 'plotlywidget'],
 ['pkg_resources', 'setuptools'],
 ['plotly', 'plotlywidget'],
 ['pkg_resources', 'setuptools'],
 ['plotly', 'plotlywidget'],
 ['pkg_resources', 'setuptools'],
 ['Cython', 'pyximport'],
 ['pkg_resources', 'setuptools'],
 ['Cython', 'pyximport'],
 ['plotly', 'plotlywidget'],
 ['pkg_resources', 'setuptools'],
 ['pkg_resources', 'setuptools'],
 ['caffe2', 'torch'],
...
```
They are under `nonunique_import_package` variable in the bottom of shared notebook.

tlegen-k:
gently ping @fridex. Would you be able to give a feedback?

fridex:
> gently ping @fridex. Would you be able to give a feedback?

As stated above, not sure if this would be a good solution. `get_or_create` logic on models creates a new entry in tables if there is no matching entry with all the attributes. Your code extends `PythonPackageVersion` which is used in other places of the project - see the adjusted part for sync_adviser_result as an example. This would mean that we eventually lose linkage to entries in the database if `import_package` will be set in some cases and in some cases it will be set to None (duplicate entries might be created).

tlegen-k:
Hi @fridex,
Here is another attempt in doing correctly.
I have created `ImportPackages` class to hold only `document_id` and `import_packages` and extended logic of `sync_solver_result` to correctly initialize instance of `ImportPackages`.

Are there any other data fields that might be beneficial to store in `ImportPackages`?

pacospace:
> > gently ping @fridex. Would you be able to give a feedback?
> 
> As stated above, not sure if this would be a good solution. `get_or_create` logic on models creates a new entry in tables if there is no matching entry with all the attributes. Your code extends `PythonPackageVersion` which is used in other places of the project - see the adjusted part for sync_adviser_result as an example. This would mean that we eventually lose linkage to entries in the database if `import_package` will be set in some cases and in some cases it will be set to None (duplicate entries might be created).

Should we make that mandatory? and resync all solvers with solver version that has that feature?

thoth/storages/graph/models.py
```diff
@@ -1688,6 +1687,18 @@ class SIAggregated(Base, BaseExtension):
     python_package_version_entity = relationship("PythonPackageVersionEntity", back_populates="si_aggregated_runs")
 
 
+class ImportPackages(Base, BaseExtension):
+    """Packages imported as a result of solver run"""
+
+    __tablename__ = "import_packages"
+
+    document_id = Column(Text, nullable=False)
```
fridex, line 1695: This can be obtained from `PythonPackageVersion` directly so no need to keep it here.

```suggestion
```

thoth/storages/graph/models.py
```diff
@@ -72,7 +72,6 @@ class PythonPackageVersion(Base, BaseExtension):
     )
     is_missing = Column(Boolean, nullable=False, default=False)
     provides_source_distro = Column(Boolean, nullable=False, default=True)
-
     # Relations
     dependencies = relationship("DependsOn", back_populates="version")
     solvers = relationship("Solved", back_populates="version")
```
fridex, line 77: You will need to add a relation to `ImportPackage`. I think you will need an association table for N-N relation here (as one package might provide the same import as another one):

https://docs.sqlalchemy.org/en/14/orm/basic_relationships.html#many-to-many

thoth/storages/graph/postgres.py
```diff
@@ -5477,6 +5477,34 @@ def sync_solver_result(self, document: dict, *, force: bool = False) -> None:
...
+                "Syncing imported packages for package %r in version %r from %r found by solver %r",
+                package_name,
+                package_version,
+                index_url,
+                solver_info,
+            )
+
+            import_packages =[]
+            for package_call in document["result"]["tree"]["packages"]:
+                if package_call.count(".") == 0 and package_call[:4] != "test"
```
fridex, line 5496: No need to have any filtering here on names. You can directly create `ImportPackage` per each entry in `document["result"]["tree"]["packages"]`.

thoth/storages/graph/models.py
```diff
@@ -1688,6 +1687,18 @@ class SIAggregated(Base, BaseExtension):
     python_package_version_entity = relationship("PythonPackageVersionEntity", back_populates="si_aggregated_runs")
 
 
+class ImportPackages(Base, BaseExtension):
```
fridex, line 1690:
```suggestion
class ImportPackage(Base, BaseExtension):
```

thoth/storages/graph/models.py
```diff
@@ -1688,6 +1687,18 @@ class SIAggregated(Base, BaseExtension):
...
+    """Packages imported as a result of solver run"""
+
+    __tablename__ = "import_packages"
+
+    document_id = Column(Text, nullable=False)
+
+    import_packages = Column(Text, nullable=False)
```
fridex, line 1697:
```suggestion
    import_package = Column(Text, nullable=False)
```

thoth/storages/graph/models.py
tlegen-k: resolved by 8a428279967604e0e603a2556370045264245cbd

thoth/storages/graph/postgres.py
tlegen-k: resolved by 8a428279967604e0e603a2556370045264245cbd

fridex:
> > > gently ping @fridex. Would you be able to give a feedback?
> > 
> > 
> > As stated above, not sure if this would be a good solution. `get_or_create` logic on models creates a new entry in tables if there is no matching entry with all the attributes. Your code extends `PythonPackageVersion` which is used in other places of the project - see the adjusted part for sync_adviser_result as an example. This would mean that we eventually lose linkage to entries in the database if `import_package` will be set in some cases and in some cases it will be set to None (duplicate entries might be created).
> 
> Should we make that mandatory? and resync all solvers with solver version that has that feature, so we don't create double entries? @fridex @tlegen-k

Making the field mandatory will not work - this information is present only in solver results and other sync methods also create `PythonPackageVersion` entries that would require a linkage to the newly introduced table.

thoth/storages/graph/postgres.py
```diff
@@ -5477,6 +5477,27 @@ def sync_solver_result(self, document: dict, *, force: bool = False) -> None:
                                 marker_evaluation_result=dependency.get("marker_evaluation_result", True),
                             )
 
+
+            # Sync imoprted package from result:tree
```
pacospace, line 5481: this should run below https://github.com/thoth-station/storages/blob/e6fbbc34f782b8555729cf9f8fe270b2753f341f/thoth/storages/graph/postgres.py#L5332 :)

thoth/storages/graph/models.py
```diff
@@ -51,6 +51,12 @@
 )
 
 
+association_table = Table('association', Base.metadata,
```
pacospace, line 54: let's follow some example we have in the databse like: https://github.com/thoth-station/storages/blob/e6fbbc34f782b8555729cf9f8fe270b2753f341f/thoth/storages/graph/models.py#L355

pacospace (commented):
Once changes are done, new alembic migration to be added to the PR :)

thoth/storages/graph/postgres.py
tlegen-k: resolved by 8e65f38da030758bc68904136a0c0888e7ed54aa

thoth/storages/graph/postgres.py
```diff
@@ -315,7 +316,7 @@ def connect(self):
         try:
             if not self.is_schema_up2date():
                 _LOGGER.debug("Database adapter connected, database is initialized")
-        except DatabaseNotInitialized as exc:
+        except Exception as exc:
```
pacospace, line 319: why this has changed?

thoth/storages/graph/models.py
```diff
@@ -50,6 +50,12 @@
     EnvironmentTypeEnum.RUNTIME.value, EnvironmentTypeEnum.BUILDTIME.value, name="environment_type", create_type=True
 )
 
+association_table = Table(
```
pacospace, line 53: this should be a class as all tables

thoth/storages/graph/postgres.py
tlegen-k: To be able to run locally on macOS. Will make as it was before sending to prod.

thoth/storages/graph/models.py
tlegen-k: logic of connection between `ImportPackage` and `PPV` was changed to Association Object as requested by using `FoundImportPackage` in commit b4147fd5152048337263037bd70f41c6a96026af

tlegen-k:
Hi @pacospace, thanks for continuous input!
I rewrote the model using association object instead of association table in b4147fd5152048337263037bd70f41c6a96026af.
Also created function `get_python_package_version_import_packages_all` that returns `ImportPackage` rows for the given `import_name`.
Next will be:
- [ ] Extend function so that: it outputs the PPV entries that are connected with the filtered 'ImportPackage`
- [ ] Create final alembic migration
Let me know what you think.

fridex:
> Hi @pacospace, thanks for continuous input!
> I rewrote the model using association object instead of association table in [b4147fd](https://github.com/thoth-station/storages/commit/b4147fd5152048337263037bd70f41c6a96026af).
> Also created function `get_python_package_version_import_packages_all` that returns `ImportPackage` rows for the given `import_name`.
> Next will be:
> 
> * [ ]  Extend function so that: it outputs the PPV entries that are connected with the filtered 'ImportPackage`
> * [ ]  Create final alembic migration
>   Let me know what you think.

Might be a good idea to also include an index on `import_package_name` given the expected size of the newly created table. 👍🏻

tlegen-k:
> > Hi @pacospace, thanks for continuous input!
> > I rewrote the model using association object instead of association table in [b4147fd](https://github.com/thoth-station/storages/commit/b4147fd5152048337263037bd70f41c6a96026af).
> > Also created function `get_python_package_version_import_packages_all` that returns `ImportPackage` rows for the given `import_name`.
> > Next will be:
> > 
> > * [ ]   Extend function so that: it outputs the PPV entries that are connected with the filtered 'ImportPackage`
> > * [ ]   Create final alembic migration
> >   Let me know what you think.
> 
> Might be a good idea to also include an index on `import_package_name` given the expected size of the newly created table. 👍🏻

Hi @pacospace and @fridex 
Added indexing in  739304e46a47c62c3c1e82c32383e086861525f8.
Final versions of `models.py` and `postgres.py`. Tested everything locally. Gives correct output.
For command `PYTHONPATH=. pipenv run python3 test_query.py` with following `test_query.py`:
```
from thoth.storages import GraphDatabase
from thoth.storages.sync import sync_solver_documents

# query part
graph = GraphDatabase()
graph.connect()


print(graph.get_python_package_version_import_packages_all("selinon"))
print(graph.get_python_package_version_import_packages_all("sklearn"))
print(graph.get_python_package_version_import_packages_all("dotenv"))
print(graph.get_python_package_version_import_packages_all("IPython"))
```
Output is: 
```
['selinon']
['scikit-learn']
['python-dotenv']
['ipython']
```
Which correctly identifies package name required to be installed.

thoth/storages/data/alembic/versions/c1839dfcdf13_add_importpackage_foundimportpackage_.py
```diff
@@ -0,0 +1,54 @@
...
+    sa.PrimaryKeyConstraint('id')
+    )
+    op.create_table('found_import_package',
+    sa.Column('python_package_version_id', sa.Integer(), nullable=False),
+    sa.Column('import_package_id', sa.Integer(), nullable=False),
+    sa.ForeignKeyConstraint(['import_package_id'], ['import_package.id'], ondelete='CASCADE'),
+    sa.ForeignKeyConstraint(['python_package_version_id'], ['python_package_version.id'], ondelete='CASCADE'),
+    sa.PrimaryKeyConstraint('python_package_version_id', 'import_package_id')
+    )
+    op.alter_column('has_symbol', 'software_environment_id',
```
pacospace, line 33: please remove line 33 to 40 from the migration manually as they are not related to this. cc @KPostOffice that encountered similar problem.

thoth/storages/graph/postgres.py
```diff
@@ -6765,3 +6790,15 @@ def get_python_package_version_all(self, analysis_document_id: str) -> List[Dict
             )
 
             return [{"package_name": i[1], "package_version": i[2], "location": i[0]} for i in query.all()]
+
+    def get_python_package_version_import_packages_all(self, import_name: str) -> str:
```
pacospace, line 6794: return type should be `List[str]`

pacospace (commented):
Few comments and please put back parts that you commented/modified for testing. Very nice work 👏🏼

thoth/storages/data/alembic/versions/c1839dfcdf13_add_importpackage_foundimportpackage_.py
tlegen-k: removed in 152184ab2cafbc52b8c8171b91a5f8317e02d60e

thoth/storages/graph/postgres.py
tlegen-k: fixed in 3f36204cb14d6f6cd76d9bdb7357c33dcfa5ad51

thoth/storages/data/alembic/versions/c1839dfcdf13_add_importpackage_foundimportpackage_.py
```diff
@@ -0,0 +1,47 @@
...
+    sa.ForeignKeyConstraint(['python_package_version_id'], ['python_package_version.id'], ondelete='CASCADE'),
+    sa.PrimaryKeyConstraint('python_package_version_id', 'import_package_id')
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(None, 'python_package_version_entity_rule', type_='unique')
+    op.alter_column('has_symbol', 'versioned_symbol_id',
```
pacospace, line 39: Also line 39 to 44 should be manually removed

tlegen-k: was removed

tlegen-k:
> Few comments and please put back parts that you commented/modified for testing. Very nice work 👏🏼

Thank you for review @pacospace!
I made sure to return part that were commented to run locally.
What should I work on next?

thoth/storages/graph/models.py
```diff
@@ -1688,6 +1687,29 @@ class SIAggregated(Base, BaseExtension):
...
+class ImportPackage(Base, BaseExtension):
+    """Packages imported as a result of solver run"""
+
+    __tablename__ = "import_package"
+
+    id = Column(Integer, primary_key=True, autoincrement=True, nullable=False)
+    import_package_name = Column(Text, nullable=True)
+
+    python_package_versions = relationship("FoundImportPackage", back_populates="import_package")
+
```
pacospace, line 1699: I actually don't see the index created for `import_package_name` as @fridex suggested. It's missing the migration also indeed, there should be something like `op.create_index`

tlegen-k: > I actually don't see the index created for `import_package_name` as @fridex suggested. It's missing the migration also indeed, there should be something like `op.create_index`

I understood as `id` field is index in line 1695:
`id = Column(Integer, primary_key=True, autoincrement=True, nullable=False)`

Screen Shot 2021-08-12 at 5 33 48 PM

Or if it is not that i can definitely work on that.

goern:
@all what is the status of this one? does it need a little love before ready to merge?

thoth/storages/graph/models.py
pacospace: See for example: https://github.com/thoth-station/storages/blob/61b9de2fa41a31547bb1b45ce7b21276455e70e7/thoth/storages/graph/models.py#L215. I think @fridex meant to have one specific for that, the ID has an index by default I think.

tlegen-k: Now clear. Implemented in d8d35af04d26d99a2f8187c78f267a6639e026a6
And created new alembic migration with adjustments to it that you pointed out previously

tlegen-k: > See for example:
> 
> https://github.com/thoth-station/storages/blob/61b9de2fa41a31547bb1b45ce7b21276455e70e7/thoth/storages/graph/models.py#L215
> . I think @fridex meant to have one specific for that, the ID has an index by default I think.

Now clear. Implemented in d8d35af04d26d99a2f8187c78f267a6639e026a6
And created new alembic migration with adjustments to it that you pointed out previously
Let me know if there is anything else)

thoth/storages/graph/postgres.py
```diff
@@ -5477,6 +5501,7 @@ def sync_solver_result(self, document: dict, *, force: bool = False) -> None:
                                 marker_evaluation_result=dependency.get("marker_evaluation_result", True),
                             )
 
+
```
fridex, line 5504:
```suggestion
```

thoth/storages/graph/postgres.py
```diff
@@ -5440,6 +5442,28 @@ def sync_solver_result(self, document: dict, *, force: bool = False) -> None:
...
+                        import_package_name=package_call,
+                    )
+
+                    FoundImportPackage.get_or_create(
+                        session,
+                        python_package_version_id=python_package_version.id,
+                        import_package_id=import_package.id,
+                    )
+
+
```
fridex, line 5466:
```suggestion
```

fridex (commented):
Nice! The code looks good to me (haven't tried it thought). Could you please squash commits to keep the git history clean? It might help in the future with debugging.

Nice work! 💯 👍🏻

tlegen-k:
> Nice! The code looks good to me (haven't tried it thought). Could you please squash commits to keep the git history clean? It might help in the future with debugging.
> 
> Nice work! 💯 👍🏻

Thank you @fridex! Have squashed 26 commits using git rebase.

pacospace:
> > Nice! The code looks good to me (haven't tried it thought). Could you please squash commits to keep the git history clean? It might help in the future with debugging.
> > Nice work! 💯 👍🏻
> 
> Thank you @fridex! Have squashed 26 commits using git rebase.

Can you please adjust to make pre-commit happy? :)

tlegen-k:
> > > Nice! The code looks good to me (haven't tried it thought). Could you please squash commits to keep the git history clean? It might help in the future with debugging.
> > > Nice work! 💯 👍🏻
> > 
> > 
> > Thank you @fridex! Have squashed 26 commits using git rebase.
> 
> Can you please adjust to make pre-commit happy? :)

Totally overlooked that one)
Thanks for pointing out @pacospace 
Fixed now.  Only needs approval labels.

pacospace:
> > > > Nice! The code looks good to me (haven't tried it thought). Could you please squash commits to keep the git history clean? It might help in the future with debugging.
> > > > Nice work! 💯 👍🏻
> > > 
> > > 
> > > Thank you @fridex! Have squashed 26 commits using git rebase.
> > 
> > 
> > Can you please adjust to make pre-commit happy? :)
> 
> Totally overlooked that one)
> Thanks for pointing out @pacospace
> Fixed now. Only needs approval labels.

Thanks!! We just miss rebase after https://github.com/thoth-station/storages/pull/2384#issuecomment-901445109 was merged and please adjust the alembic migration number now, because that PR already introduced a migration :)

thoth/storages/graph/postgres.py
```diff
@@ -6786,3 +6809,15 @@ def get_python_package_version_all(self, analysis_document_id: str) -> List[Dict
...
+        """Retrieve Python package name for the given import package name."""
+        with self._session_scope() as session:
+            query = (
+                session.query(PythonPackageVersion)
+                .join(FoundImportPackage)
+                .join(ImportPackage)
+                .filter(ImportPackage.import_package_name == import_name)
+                .filter(PythonPackageVersion.id == FoundImportPackage.python_package_version_id)
+            )
+            return [i.package_name for i in query.all()]
```
fridex, line 6823: Can the query return `package_name`, `package_version`, `index_url` triplet so the package is uniquely identified?

fridex (approved):
lgtm, thanks @tlegen-k 💯 👍🏻

tlegen-k:
> lgtm, thanks @tlegen-k 💯 👍🏻

Thank you for helping!

PR #2420, opened 2021-08-17
Improve performance of the unsolved Python packages query

fridex:

- [x] No

- [x] Yes

As observed during ingestion, the performance of unsolved query used in graph-refresh-job significantly dropped with more data ingested. After digging into this, the issue seems to be in the rules association that resulted in a sequential scan across the whole table of rules (that has few thousand entries) for each package we ingested. Naturally, this grows significantly. The optimization uses LEFT OUTER JOIN to speed up the query. I tested this query in the stage environment - currently, the query finishes in ~3 seconds (in opposite to hours).


fridex:
I will merge this to unblock ingestion in the stage. Feel free to raise any comments.

/approve

PR #2468, opened 2021-11-03
WIP Implementation of store_old_documents() for purge-job

mayaCostantini:

Related to https://github.com/thoth-station/thoth-application/issues/1516

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

A new method to store old Thoth solver documents as requested by the purge-job issue mentioned above.

## Description

The store_old_documents() method has been added to the GraphDatabase class in thoth/storage/graph/postgres.py to handle the storage of solver, adviser and package-extract documents about to be deleted during Thoth's purge job.


thoth/storages/graph/postgres.py
```diff
@@ -6710,6 +6710,45 @@ def get_bloat_data(self) -> List[Dict[str, Any]]:
...
+            document_id_requirements = tuple(document_id.split("-")[1:4])
+
+            if document_id_requirements == (os_name, os_version, python_version):
+                json_document = solver_store.retrieve_document(document_id)
+                solver_store = store_document(json_document)
+                _LOGGER.info(f"Stored to be deleted document {document_id} in Ceph")
+
+    def store_old_adviser_documents(self, end_datetime=None, adviser_version=None):
+        """Stores old adviser documents to be deleted in Ceph"""
+        solver_store = SolverResultsStore()
```
fridex, line 6728: Copy-pasta - also for the package-extract docs below:

```suggestion
        adviser_store = AdviserResultsStore()
```

thoth/storages/graph/postgres.py
```diff
@@ -6710,6 +6710,45 @@ def get_bloat_data(self) -> List[Dict[str, Any]]:
...
+        solver_store.connect()
+
+        for document_id in solver_store.get_document_listing():
+            document_id_requirements = tuple(document_id.split("-")[1:4])
```
fridex, line 6732: The listing for adviser docs is somehow unique - there are stored also serialized requests (see docs with `.request`). This will need to be skipped.

thoth/storages/graph/postgres.py
```diff
@@ -6710,6 +6710,45 @@ def get_bloat_data(self) -> List[Dict[str, Any]]:
 
         return self._process_bloat_data_results(tables=tables)
 
+    def store_old_solver_documents(self, os_name=None, os_version=None, python_version=None):
```
fridex, line 6713: It might be a good idea to make these methods more generic - move them to the result base which would accept instantiated `GraphDabase` to query document ids and move them to a new location defined by a prefix (a CephStore can be instantiated within the implemented method with the specified prefix).

fridex (commented):
Nice work! Few comments left, we can sync on a call if you want. The approach looks correct though 👍🏻

thoth/storages/graph/postgres.py
mayaCostantini: Thanks, it is now fixed by : [Corrected value of adviser_store to AdviserResultsStore() at L6728](https://github.com/thoth-station/storages/pull/2468/commits/16a818dc4e8961111dcaa06f6d800767963935e0)

thoth/storages/solvers.py
```diff
@@ -29,3 +29,16 @@ class SolverResultsStore(ResultStorageBase):
...
+        solver_document_ids = graph.get_solver_run_document_ids_all(
+            os_version=os_version, os_name=os_name, python_version=python_version
+        )
+        self.connect()
+
+        for solver_document_id in solver_document_ids:
+            solver_document_id_requirements = tuple(solver_document_id.split("-")[1:4])
+            if solver_document_id_requirements == (os_name, os_version, python_version):
+                json_document = self.ceph.retrieve_document(document_id=solver_document_id)
+                self.ceph.store_document(document=json_document, document_id=solver_document_id)
```
fridex, line 44: Would this code store `json_document` in a new location or in will it overwrite already existing document with the same content?

mayaCostantini: The `store_document` method used here is actually based on the `boto3` `Object().put` method , were the document name to be stored is specified by the second argument of the `Object` constructor, formed by the prefix chosen in `purge-job/app.py` (for example with `solver_store = SolverResultsStore(prefix="old-solver-documents")`). From what I understand, unless using [bucket versioning](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Versioning.html), if the file with name `"prefix" + filename"` is already present in the bucket, this operation will overwrite it.

fridex: Let's say we want to run the purge job and make sure the old documents are moved to a new location. Pseudocode to do that (haven't tested, just and idea how to do that):

```python
from thoth.storages import GraphDatabase
from thoth.storages import CephStore
from thoth.storages import SolverResultsStore
from thoth.common import datetime2datetime_str

graph = GraphDatabase()
graph.connect()

solver_store = SolverResultsStore()
solver_store.connect()

# Move to solver-purge-<timestamp>.
target_prefix = f"{solver_store.ceph.prefix.rsplit("/", maxsplit=1)[0]}/solver-purge-{datetime2datetime_str()}/"
target_store = CephStore(prefix=target_prefix)
target_store.connect()

for solver_document_id in solver_document_ids:
    document = solver_store.retrieve_document(document_id=solver_document_id)
    target_store.store_document(document, document_id=solver_document_id)
    solver_store.delete_document(document_id=solver_document_id)  # needs to be implemented in ceph.py
    graph.remove_solver_entry(document_id=solver_document_id)  # needs to be implemented in postgres.py
```

This logic can be refactored and placed to `GraphDatabase` in postgres.py which would accept `solver_store` and `target_prefix` to move documents based on the query and delete the entry in the database (returns number of documents purged so we can eventually track it):

```python
class GraphDatabase:

    def purge_solver_documents(
        solver_store: SolverResultsStore,
        target_prefix: str,
        *,
        os_name: Optional[str] = None,
        Optional[os_version] = None,
        python_version: Optional[str] = None
    ) -> int:
        ...
```

Would this solution make sense?

mayaCostantini: It makes sense, thank you!

thoth/storages/graph/postgres.py
```diff
@@ -6751,6 +6753,32 @@ def delete_solved(self, *, os_name: str, os_version: str, python_version: str) -
...
+        solver_document_ids = self.get_solver_run_document_ids_all(
+            os_version=os_version, os_name=os_name, python_version=python_version
+        )
+
+        deleted_solver_documents_count = 0
+
+        for solver_document_id in solver_document_ids:
+            document = solver_store.retrieve_document(document_id=solver_document_id)
+            target_store.store_document(document, document_id=solver_document_id)
+            solver_store.delete(object_key=solver_document_id)
```
fridex, line 6777: It looks like `SolverResultsStore` has no `delete` method. It might be a good idea to provide one or call `solver_store.ceph.delete` instead:

```suggestion
            solver_store.ceph.delete(object_key=solver_document_id)
```

thoth/storages/result_base.py
```diff
@@ -169,6 +169,10 @@ def store_document(self, document: dict) -> str:
         self.ceph.store_document(document, document_id)
         return document_id
 
+    def store_old_documents(self, graph):
```
fridex, line 172: This might be eventually removed.

thoth/storages/graph/postgres.py
```diff
@@ -6751,6 +6753,32 @@ def delete_solved(self, *, os_name: str, os_version: str, python_version: str) -
                 .delete()
             )
 
+    def purge_solver_documents(
+        self, os_name: Optional[str] = None, os_version: Optional[str] = None, python_version: Optional[str] = None
```
fridex, line 6757:
```suggestion
        self, *, os_name: Optional[str] = None, os_version: Optional[str] = None, python_version: Optional[str] = None
```

fridex (commented):
Changes look good to me 👍🏻 Just a few minor comments.

PR #2477, opened 2021-11-15
Use explicit exports

fridex:

mypy can complain about this:

```
thoth/adviser/digests_fetcher.py:25:1: error: Module "thoth.storages" does not explicitly export attribute "GraphDatabase"; implicit reexport disabled  [attr-defined]
```

- [x] No

- [x] No


harshad16 (approved):
lgtm
thanks :100:

PR #2482, opened 2021-11-18
Add methods to purge old documents and store them on Ceph

mayaCostantini:

Related to [#1516 on thoth-application](https://github.com/thoth-station/thoth-application/issues/1516)

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

Add methods to purge old adviser, solver and package extract documents and to store them to Ceph for future use, that will match the commands implemented on the purge-job repo.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -6751,7 +6751,6 @@ class GraphDatabase(SQLBase):
         self, *, os_name: Optional[str] = None, os_version: Optional[str] = None, python_version: Optional[str] = None
     ) -> int:
         """Store and purge to be deleted solver documents to Ceph"""
-        self.connect()
         solver_store = SolverResultsStore()
         solver_store.connect()
 
@@ -6777,7 +6776,6 @@ class GraphDatabase(SQLBase):
         self, *, end_datetime: Optional[str] = None, adviser_version: Optional[str] = None
     ) -> int:
         """Store and purge to be deleted adviser documents to Ceph"""
-        self.connect()
         adviser_store = AdvisersResultsStore()
         adviser_store.connect()
 
@@ -6814,7 +6812,6 @@ class GraphDatabase(SQLBase):
         self, *, end_datetime: Optional[str] = None, package_extract_version: Optional[str] = None
     ) -> int:
         """Store and purge to be deleted package extract documents to Ceph"""
-        self.connect()
         package_extract_store = AnalysisResultsStore()
         package_extract_store.connect()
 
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -6738,6 +6740,108 @@ def get_bloat_data(self) -> List[Dict[str, Any]]:
...
+        for solver_document_id in solver_document_ids:
+            document = solver_store.retrieve_document(document_id=solver_document_id)
+            target_store.store_document(document, document_id=solver_document_id)
+            solver_store.ceph.delete(object_key=solver_document_id)
+            deleted_solver_documents_count += self.delete_solver_result(solver_document_id=solver_document_id)
+
+        return deleted_solver_documents_count
+
+    def purge_adviser_documents(
+        self, *, end_datetime: Optional[str] = None, adviser_version: Optional[str] = None
```
fridex, line 6770: Looking into this, should `end_datetime` be of type `datetime`? Otherwise, it is confusing. I see this also in other parts of the module - we should fix it. Let's keep this as-is for now and adjust it in a separate PR.

thoth/storages/graph/postgres.py
```diff
@@ -6738,6 +6740,108 @@ def get_bloat_data(self) -> List[Dict[str, Any]]:
 
         return self._process_bloat_data_results(tables=tables)
 
+    def purge_solver_documents(
+        self, *, os_name: Optional[str] = None, os_version: Optional[str] = None, python_version: Optional[str] = None
+    ) -> int:
+        """Store and purge to be deleted solver documents to Ceph"""
+        self.connect()
```
fridex, line 6747: The convention is that a user of this adapter makes sure the adapter is connected:

```suggestion
```

Here and other methods as well.

fridex (commented):
Looks really good. Few cosmetic comments. Otherwise, we could give this a try in the test environment. 👍🏻

thoth/storages/graph/postgres.py
mayaCostantini: Thanks, I removed it!

thoth/storages/graph/postgres.py
mayaCostantini: You are right, I typed it as a string because I forgot that the actual argument from `purge-job` uses the `parse` method from `dateutil`, I will open a PR and correct this !

fridex (approved):
lgtm, thanks! 👍🏻 nice work

mayaCostantini:
> lgtm, thanks! 👍🏻 nice work

Thank you!

PR #2488, opened 2021-11-24
Allow querying module imports

fridex:

- [x] No

- [ ] Yes
- [x] No

## Description

With this change, we will be able to query imports (prefix query). See examples for more info.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -7050,25 +7050,58 @@ class GraphDatabase(SQLBase):
     def get_python_package_version_import_packages_all(
         self, import_name: str, distinct: bool = False
     ) -> List[Dict[str, str]]:
-        """Retrieve Python package name for the given import package name."""
+        """Retrieve Python package name for the given import package name.
+
+        Examples:
+        >>> from thoth.storages import GraphDatabase
+        >>> graph = GraphDatabase()
+        >>> graph.get_python_package_version_import_packages_all("faust.*")
+        [
+           {'import': 'faust.web.apps',
+            'index_url': 'https://pypi.org/simple',
+            'package_name': 'faust',
+            'package_version': '1.9.0'},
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -7050,25 +7050,59 @@ def get_python_package_version_all(self, analysis_document_id: str) -> List[Dict
...
             query = (
-                session.query(PythonPackageVersion)
-                .join(FoundImportPackage)
-                .join(ImportPackage)
-                .filter(ImportPackage.import_package_name == import_name)
-                .filter(PythonPackageVersion.id == FoundImportPackage.python_package_version_id)
+                query.filter(PythonPackageVersion.id == FoundImportPackage.python_package_version_id)
                 .join(PythonPackageIndex)
                 .filter(PythonPackageVersion.python_package_index_id == PythonPackageIndex.id)
+                .distinct()
```
pacospace, line 7091: do we introduce it by default?

pacospace (commented):
One comment, otherwise lgtm! 💯

thoth/storages/graph/postgres.py
fridex: I'll remove it as the parameter is already propagated on user api. BTW not sure about the applicability of `distinct` here - the database should keep each record unique anyway.

fridex:
Adjusted. PTAL @pacospace

thoth/storages/graph/postgres.py
pacospace: > 

the distinct is because of the solver, we receive twice the same package otherwise: https://github.com/thoth-station/storages/blob/0b0e38359369efdb3e6547eba3debf48b1c25d39/thoth/storages/graph/models.py#L64

pacospace (approved):
Thanks! 💯

PR #2499, opened 2021-11-30
Fix v0.61.0 changelog computed

fridex:

- [x] No

- [x] No

## Description

This was caused by a wrong tag created (`kebechet-` prefix issue).


harshad16 (approved):
lgtm
thanks :100:

PR #2512, opened 2021-12-21
Extend get_software_environments_all query to provide also package_extract_run id and datetime.

pacospace:

Related-To: https://github.com/thoth-station/user-api/issues/1543

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -2813,6 +2813,32 @@ class GraphDatabase(SQLBase):
         """Get software environments (external or internal) registered in the graph database."""
         os_name = map_os_name(os_name)
         os_version = normalize_os_version(os_name, os_version)
+        """Get software environments (external or internal) registered in the graph database.
+
+        Examples:
+        >>> from thoth.storages import GraphDatabase
+        >>> graph = GraphDatabase()
+        >>> graph.get_software_environments_all()
+
+        [
+            {
+                'cuda_version': None,
+                'datetime': datetime.datetime(2021, 12, 15, 19, 27, 52, 803266),
+                'env_image_name': None,
+                'env_image_tag': None,
+                'environment_name': 'quay.io/thoth-station/s2i-thoth-ubi8-py39:v0.32.3',
+                'environment_type': 'RUNTIME',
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -2799,6 +2799,7 @@ def get_hardware_environments_all(
     def get_software_environments_all(
         self,
         is_external: bool = False,
+        include_ids: bool = False,
```
fridex, line 2802: Do we use this query elsewhere which requires keeping the API strictly backward compatible?

pacospace: No to my knowledge 🤔 should we make it default choice?

pacospace: Maybe metrics-exporter

fridex: I did find it used in metrics-exporter. I think we can safely just extend the dictionary returned with additional values introduced - no need to add `include_ids` parameter.

pacospace: > Maybe metrics-exporter

ok no, we are good

pacospace: > I did find it used in metrics-exporter. I think we can safely just extend the dictionary returned with additional values introduced - no need to add `include_ids` parameter.

Adjusted, thanks!!

thoth/storages/graph/postgres.py
```diff
@@ -2842,8 +2868,49 @@ def get_software_environments_all(
...
+                        "cuda_version": r[0],
+                        "datetime": r[12],
+                        "env_image_name": r[1],
+                        "env_image_tag": r[2],
+                        "environment_name": r[3],
+                        "environment_type": r[4],
+                        "image_sha": r[5],
+                        "os_name": r[6],
+                        "os_version": r[7],
+                        "package_extract_run_id": r[11],
```
fridex, line 2907:
```suggestion
                        "package_extract_document_id": r[11],
```

fridex (commented):
One minor comment, otherwise lgtm 👍🏻

fridex (approved):
lgtm, thanks! 💯 👍🏻

PR #2518, opened 2021-12-21
Do not propagate index url when asking for package names

fridex:

- [x] No


fridex:
Looks like this query is used in https://github.com/thoth-station/package-update-job/blob/81426813eed6d289325a3a1593c21ecc192a9990/producer.py#L193 We can create a new one for the desired functionality.

/close

PR #2519, opened 2021-12-21
Fix return type

fridex:

- [x] No

- [x] No

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -5008,7 +5008,7 @@ class GraphDatabase(SQLBase):
 
         @params date_: DD-MM-YY
         """
-        return datetime.strptime(date_, "%d-%m-%Y").date()
+        return datetime.strptime(date_, "%d-%m-%Y")
 
     def get_adviser_run_document_ids_all(
         self,


fridex:
/assign @pacospace 

PTAL

thoth/storages/graph/postgres.py
```diff
@@ -5003,7 +5004,7 @@ def get_adviser_run_count_per_source_type(
         return self._filter_source_type(results=results)
 
     @staticmethod
-    def _create_date_filter(date_: str) -> datetime:
+    def _create_date_filter(date_: str) -> date:
```
pacospace, line 5007: Actually based on how we use this method in the library, it has to output datetime, maybe we can remove `date()`, because we should not (cannot, I guess it fails when done in Python) compare datetime and date: https://github.com/thoth-station/storages/blob/11fa9910270780ff02aa3be28d328e3410fc1b4f/thoth/storages/graph/postgres.py#L5041

wdyt?

fridex: That's true 👍🏻 Shouldn't the current code raise exceptions when comparing date and datetime? The PR has been adjusted.

pacospace: I think only in Python the exception is raised
```
Traceback (most recent call last):
  File "mix_test.py", line 16, in <module>
    if date_ < today:
TypeError: can't compare datetime.datetime to datetime.date
```

but there we are comparing sqlalchemy object column datetime, so maybe they don't have these checks there :/

pacospace (approved):
trhanks! 🍻

PR #2522, opened 2021-12-21
Introduce query for listing solved Python package names

fridex:

Related: https://github.com/thoth-station/user-api/pull/1548

- [x] No

- [x] No

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -814,6 +814,49 @@ class GraphDatabase(SQLBase):
         os_version = normalize_os_version(os_name, os_version)
         return self.__class__.get_python_packages_all(**locals())
 
+    def get_solved_python_package_names_all(
+        self,
+        *,
+        start_offset: int = 0,
+        count: Optional[int] = DEFAULT_COUNT,
+        os_name: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


fridex:
/assign @pacospace 

PTAL

thoth/storages/graph/postgres.py
```diff
@@ -814,6 +814,49 @@ def get_solved_python_packages_all(
         os_version = normalize_os_version(os_name, os_version)
         return self.__class__.get_python_packages_all(**locals())
 
+    def get_solved_python_package_names_all(
```
pacospace, line 817: I think we have the same already: https://github.com/thoth-station/storages/blob/11fa9910270780ff02aa3be28d328e3410fc1b4f/thoth/storages/graph/postgres.py#L2961 but called `get_python_package_version_names_all` because `python_package_version` is synonym of `solved` for us.

pacospace: my bad if confusing, I realized we have mixed nomenclature, sometimes with solved, sometimes without

fridex: Oh, thanks for pointing that. Makes life easier 👍🏻 

/close

PR #2525, opened 2021-12-22
Introduce query for obtaining solver documents for the given package

fridex:

- [x] No

- [x] No

## Description

Obtain solver documents based on the supplied python package.


fridex:
/assign @pacospace 

PTAL

pacospace (approved):
Thanks 💯 lgtm!!

PR #2532, opened 2022-01-04
Provide a method that counts number of Python package names

fridex:

- [x] No

- [x] No

## Description

Implement this method to properly obtain the number of solved package names. The suggested method in https://github.com/thoth-station/storages/pull/2522#discussion_r773454447 does not work here because it takes into account package version and index url as well.


pacospace (approved):
lgtm! 💯

PR #2543, opened 2022-01-18
add break statement and remove unique slug constraint

KPostOffice:


thoth/storages/data/alembic/versions/c427c501f966_remove_unique_constraint_from_slug_in_.py
```diff
@@ -0,0 +1,30 @@
...
+revision = 'c427c501f966'
+down_revision = 'b718c0f8bc53'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint('kebechet_github_installations_slug_key', 'kebechet_github_installations', type_='unique')
+    op.create_unique_constraint(None, 'python_package_version_entity_rule', ['id'])
```
pacospace, line 22: I think this should not be part of the migration right?

KPostOffice: I think it needs to be added because the table needs to have a unique constraint somewhere. I think this effectively changes the primary key from the slug to the id. I might be able to force a unique constraint between slug and environment name instead.

pacospace: Sorry @KPostOffice my comment is was related to `python_package_version_entity_rule` which is not part of the changes you made, that is why I said it should not be part. It was probably created by the sqlalchemy for some past not tracked changes.

Regarding the unique constraint, isn't it already part of https://github.com/thoth-station/storages/blob/dbfc04cc2cb780bb21b36edcd607e67839a86a36/thoth/storages/graph/models.py#L1601?

pacospace (approved):
lgtm! 💯

pacospace:
@KPostOffice please take care of pre-commit

PR #2569, opened 2022-02-01
Retrieve datetime of last synced solver

mayaCostantini:

Related to https://github.com/thoth-station/metrics-exporter/issues/822

- No

- Yes

Implement a query to retrieve the datetime of the last solver synced in the database (see first acceptance criteria of https://github.com/thoth-station/metrics-exporter/issues/822)

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -407,25 +407,24 @@ class GraphDatabase(SQLBase):
             return result[0]
 
     def get_last_solver_datetime(
-        self, os_name: Optional[str], os_version: Optional[str], python_version: Optional[str]
+        self, os_name: Optional[str] = None, os_version: Optional[str] = None, python_version: Optional[str] = None
     ) -> datetime:
         """Get the datetime of the last solver run synced in the database."""
         with self._session_scope() as session:
-            result = session.query(func.max(Solved.datetime))
-
+            result = session.query(Solved.datetime)
             if os_name or os_version or python_version:
                 result = result.filter(Solved.ecosystem_solver_id == EcosystemSolver.id)
 
-            if os_name is not None:
-                result = result.filter(EcosystemSolver.os_name == os_name)
+                if os_name is not None:
+                    result = result.filter(EcosystemSolver.os_name == os_name)
 
-            if os_version is not None:
-                result = result.filter(EcosystemSolver.os_version == os_version)
+                if os_version is not None:
+                    result = result.filter(EcosystemSolver.os_version == os_version)
 
-            if python_version is not None:
-                result = result.filter(EcosystemSolver.python_version == python_version)
+                if python_version is not None:
+                    result = result.filter(EcosystemSolver.python_version == python_version)
 
-            return result.first()[0]
+        return max([datetime[0].isoformat() for datetime in result.all()])
 
     @staticmethod
     def normalize_python_package_name(package_name: str) -> str:
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -406,6 +406,14 @@ class GraphDatabase(SQLBase):
             result = session.execute(query).fetchone()
             return result[0]
 
+    def get_last_solver_datetime(self) -> datetime:
+        """Get the datetime of the last solver run synced in the database."""
+        query = "SELECT datetime FROM solver WHERE datetime IN (SELECT max(datetime) FROM solver)"
+
+        with self._session_scope() as session:
+            result = session.execute(query).fetchone()
+            return result[0]
+
     @staticmethod
     def normalize_python_package_name(package_name: str) -> str:
         """Normalize Python package name based on PEP-0503."""
@@ -7107,7 +7115,7 @@ class GraphDatabase(SQLBase):
     def purge_solver_documents(
         self, *, os_name: Optional[str] = None, os_version: Optional[str] = None, python_version: Optional[str] = None
     ) -> int:
-        """Store and purge to be deleted solver documents to Ceph"""
+        """Store and purge to be deleted solver documents to Ceph."""
         solver_store = SolverResultsStore()
         solver_store.connect()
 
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -406,6 +406,14 @@ def get_alembic_version_count_all(self) -> int:
             result = session.execute(query).fetchone()
             return result[0]
 
+    def get_last_solver_datetime(self) -> datetime:
+        """Get the datetime of the last solver run synced in the database."""
+        query = "SELECT datetime FROM solver WHERE datetime IN (SELECT max(datetime) FROM solver)"
```
fridex, line 411: Could this be rewritten to an SQLAlchemy query? It should be possible, something like (haven't tried it):

```python
from sqlalchemy.sql.expression import func

    result = session.query(func.max(Solver.datetime))
# ...
```

mayaCostantini: Sure, thanks! :+1:

thoth/storages/graph/postgres.py
```diff
@@ -406,6 +406,12 @@ def get_alembic_version_count_all(self) -> int:
             result = session.execute(query).fetchone()
             return result[0]
 
+    def get_last_solver_datetime(self) -> datetime:
```
pacospace, line 409: do we also need solver filters (os_name, os_version, python_version)? or only generic latest solver run?

mayaCostantini: That's a good question. Given the issue https://github.com/thoth-station/metrics-exporter/issues/822 I understood this was for all solver documents, @fridex wdyt?

pacospace: actually, can't we reuse this query which has all filters already https://github.com/thoth-station/storages/blob/9bbc64b06c11efe7f85f616aea76be5db6f326f8/thoth/storages/graph/postgres.py#L2540?

mayaCostantini: I agree, this query has a parameter to sort the documents in datetime order (latest first), so it would be more convenient :+1:

fridex: > do we also need solver filters (os_name, os_version, python_version)? or only generic latest solver run?

We wanted to expose datetime in dashboards so we are aware "how old the database is". Additional filters can be introduced to the method for completeness, but I don't know if that's something we want to show on dashboards. Only a generic one could be good enough.

> actually, can't we reuse this query which has all filters already

The method is supposed to get the latest datetime for solvers. The referenced one will get solver ids (which might require an additional query to get the actual datetime).

mayaCostantini: > The method is supposed to get the latest datetime for solvers.

Right, then it's better to leave it as a different query :+1: 
Concerning the filters, I will let the query as it is if it's just a matter of showing the latest update of the database

fridex: > I will let the query as it is if it's just a matter of showing the latest update of the database

Up to you. I can imagine a more detailed dashboard overview showing datetime for each solver registered and freshness of the dependency data it computes - that might uncover issues that certain runtime environment is not updating (ex. some deployment issue).

mayaCostantini: I will leave this as it is for the moment, and we can open a separate issue to improve the dashboard with more details if that's ok with you :+1:

thoth/storages/graph/postgres.py
```diff
@@ -406,6 +406,27 @@ def get_alembic_version_count_all(self) -> int:
             result = session.execute(query).fetchone()
             return result[0]
 
+    def get_last_solver_datetime(
+        self, os_name: Optional[str], os_version: Optional[str], python_version: Optional[str]
```
pacospace, line 410: `Optional[str] = None` to be optional

mayaCostantini:
I made corrections to the query and it works locally, PTAL @fridex @pacospace :+1:

thoth/storages/graph/postgres.py
```diff
@@ -406,6 +406,26 @@ def get_alembic_version_count_all(self) -> int:
             result = session.execute(query).fetchone()
             return result[0]
 
+    def get_last_solver_datetime(
+        self, os_name: Optional[str] = None, os_version: Optional[str] = None, python_version: Optional[str] = None
+    ) -> datetime:
```
pacospace, line 411: I tested locally and this is a `string` actually, we need to modify the output if you expect to be of type `datetime`

mayaCostantini: Right, thanks!

pacospace: But if this method is going to be used in metrics-exporter, string is totally fine :)

mayaCostantini: If it's ok with you, we can leave it as a `datetime` just in case we reuse the method somewhere else, and I can eventually make changes later if it is more practical to have a string :)

pacospace: fine for me, but I'm not mistaken we will need to reconvert it to string to use it as label in a prometheus metric :)

thoth/storages/graph/postgres.py
```diff
@@ -406,6 +406,26 @@ def get_alembic_version_count_all(self) -> int:
...
+                if os_name is not None:
+                    result = result.filter(EcosystemSolver.os_name == os_name)
+
+                if os_version is not None:
+                    result = result.filter(EcosystemSolver.os_version == os_version)
+
+                if python_version is not None:
+                    result = result.filter(EcosystemSolver.python_version == python_version)
+
+        return datetime.strptime(max([datetime[0].isoformat() for datetime in result.all()]), "%Y-%m-%d %H:%M:%S.%f")
```
pacospace, line 427: `ValueError: time data '2022-02-06T23:08:06.688200' does not match format '%Y-%m-%d %H:%M:%S.%f'`

pacospace (approved):
Thanks 💯

PR #2575, opened 2022-02-02
Fix container image query

mayaCostantini:

Related to https://github.com/thoth-station/user-api/issues/1645

- No

- Yes

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -3039,9 +3039,8 @@ class GraphDatabase(SQLBase):
 
         if package_name:
             query = (
-                query.join(PackageExtractRun)
-                .join(Identified)
-                .join(PythonPackageVersionEntity)
+                query.filter(PackageExtractRun.id == Identified.package_extract_run_id)
+                .filter(Identified.python_package_version_entity_id == PythonPackageVersionEntity.id)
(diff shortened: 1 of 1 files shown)


harshad16:
lgtm 
thanks :100: 

/hold
Testing

harshad16:
This has failed:
```
"sqlalchemy.exc.AmbiguousForeignKeysError: Can't determine join between 'Join object on Join object on Join object on software_environment(140034398813248) and package_extract_run(140034399825296)(140034076348320) and identified(140034398967456)(140034076348224) and python_package_version_entity(140034400171008)' and 'package_extract_run'; tables have more than one foreign key constraint relationship between them. Please specify the 'onclause' of this join explicitly.\n
```

harshad16:
however when i run manually this SQL statement
```
SELECT * FROM software_environment JOIN package_extract_run ON software_environment.id = package_extract_run.software_environment_id JOIN identified ON package_extract_run.id =identified.package_extract_run_id JOIN python_package_version_entity ON  python_package_version_entity.id = identified.python_package_version_entity_id
```
it provide a complete join table

harshad16:
we might have to change the filter statement. 
@mayaCostantini maybe the other query statement would be better fit here.
where join happen with filter condition

harshad16 (approved):
lgtm
The issue was tested this solves the issue.
thanks :100:

PR #2580, opened 2022-02-04
Fix Ceph variable names and add examples

fridex:

- [x] No

- [x] No


pacospace (approved):
Thanks! lgtm 💯

PR #2583, opened 2022-02-07
Fix adapter name for package-extract documents

mayaCostantini:

Related to https://github.com/thoth-station/purge-job/issues/31

- No

- Yes


thoth/storages/graph/postgres.py
```diff
@@ -7169,7 +7170,7 @@ def purge_package_extract_documents(
         self, *, end_datetime: Optional[datetime] = None, package_extract_version: Optional[str] = None
     ) -> int:
         """Store and purge to be deleted package extract documents to Ceph"""
-        package_extract_store = AnalysisResultsStore()
+        package_extract_store = PackageExtractResultsStore()
         package_extract_store.connect()
 
         target_prefix = f"{package_extract_store.ceph.prefix.rsplit('/', maxsplit=1)[0]}/package-extract-purge-{datetime2datetime_str()}/"
```
fridex, line 7176: `AnalysisResultsStore` is somehow historically named the way it is named - I agree it is inconsistency we carry in sources. A lot of code relies on the name now, so might be a good idea to keep it as is.

The issue here could be the prefix calculated - 

```
data/ocp4-stage/analysis/package-extract-purge-<timestamp>/
```

it looks to me the analysis part is not split correctly. Maybe something like this could do the trick:

```suggestion
        target_prefix = f"{package_extract_store.ceph.prefix.rsplit('/', maxsplit=2)[0]}/package-extract-purge-{datetime2datetime_str()}/"
```

I suspect the prefix ends with `/` and hence the `analysis` part is untouched. But worth verifying - here and in other purge related methods.

mayaCostantini: Let's keep `AnalysisResultsStore` then :+1: 
I see the issue with the prefix name, and I think your change should solve the issue and ignore the last element of the array which is the name of the adapter (i.e. `analysis` here). I will also make the change for other `purge_*` methods

fridex (approved):
lgtm, thanks! 💯 👍🏻

PR #2595, opened 2022-02-11
Increase max_wal_size in postgres config

mayaCostantini:

- No

- No

Increase the `max_wal_size` parameter in postgres config to 2GB to avoid error linked to checkpoints occurring too frequently.


fridex (approved):
lgtm

Let's see if this will work properly. We could eventually revert this anytime, it is a local run. 👍🏻 

Thanks! 💯

PR #2602, opened 2022-02-24
update metadata.id instead of passing it to 'create_ppv'

KPostOffice:

https://github.com/thoth-station/integration-tests/issues/250

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

When new solver docs are synced they can update package metadata resulting in a new `metadata` row with a unique id. When this is passed to `_get_or_create` along with the other values an IntegrityError is raised because the other params already define a unique entry. However, no entry with the exact **kwargs is found so an error is raised. In this PR I removed the package_metadata_id from the original `_create...` query and then added it using an update statement below.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -4761,8 +4761,11 @@ class GraphDatabase(SQLBase):
             os_version=os_version,
             python_version=python_version,
             entity_id=entity.id,
-            python_package_metadata_id=python_package_metadata_id,
         )
+
+        # including this value in "get_or_create" will cause errors because it is not part of a unique entry
+        if python_package_metadata_id:
+            python_package_version.python_package_metadata_id = python_package_metadata_id
 
         return python_package_version
 


fridex (commented):
Nice! This looks like a very elegant solution. Have you tested it? (I don't see any direct blockers though) 👍🏻 💯

KPostOffice:
/hold
@fridex have not tested it yet. Not sure the best way to go about it

fridex:
> Not sure the best way to go about it

Trying to sync a solver document with results multiple times. The second sync can have metadata adjusted (manually in the JSON file should work fine).

KPostOffice:
So I did this but it still had the same issue because the default for `metadata` id is `None`. I moved the new logic into the called function

KPostOffice:
@fridex PTAL

thoth/storages/graph/postgres.py
```diff
@@ -4761,9 +4761,13 @@ def _create_python_package_version(
             os_version=os_version,
             python_version=python_version,
             entity_id=entity.id,
-            python_package_metadata_id=python_package_metadata_id,
         )
 
+        # including this value in "get_or_create" will cause errors because it is not part of a unique entry
+        if python_package_metadata_id:
```
fridex, line 4767:
```suggestion
        if python_package_metadata_id is not None:
```

fridex (commented):
Looks good to me. @pacospace wdyt?

thoth/storages/graph/postgres.py
```diff
@@ -4761,9 +4761,13 @@ def _create_python_package_version(
             os_version=os_version,
             python_version=python_version,
             entity_id=entity.id,
-            python_package_metadata_id=python_package_metadata_id,
         )
 
+        # including this value in "get_or_create" will cause errors because it is not part of a unique entry
+        if python_package_metadata_id is not None:
+            python_package_version.python_package_metadata_id = package_metadata.id
```
pacospace, line 4768: `package_metadata` is not passed to this method

thoth/storages/graph/postgres.py
```diff
@@ -6030,7 +6034,6 @@ def sync_solver_result(self, document: dict, *, force: bool = False) -> None:
                         os_version=ecosystem_solver.os_version,
```
pacospace, line 6034: `package_metadata` needs to be provided here

pacospace (commented):
I was trying to test and there is one parameter missing.

thoth/storages/graph/postgres.py
KPostOffice: ty

thoth/storages/graph/postgres.py
KPostOffice: Oh, sorry. I misread

pacospace: no problem @KPostOffice, now I see what you meant, you were referring to the id of package_metadata. All good now :)

thoth/storages/graph/postgres.py
```diff
@@ -4761,9 +4761,13 @@ def _create_python_package_version(
             os_version=os_version,
             python_version=python_version,
             entity_id=entity.id,
-            python_package_metadata_id=python_package_metadata_id,
         )
 
+        # including this value in "get_or_create" will cause errors because it is not part of a unique entry
+        if python_package_metadata_id:
+            python_package_version.python_package_metadata_id = python_package_metadata_id
+            session.commit()
```
pacospace, line 4769: this commit cannot be done here because this method is used in other methods so there are still opreations to be done later. for example it is used during solver sync method (in particular to create the solved record in the Solved table: https://github.com/thoth-station/storages/blob/feec955fb80f84006659662bad7916520de7907f/thoth/storages/graph/postgres.py#L6179).

Testing you get the following error:

```
Traceback (most recent call last):
  File "mix_test.py", line 42, in <module>
    output = sync_solver_documents(
  File "/home/fmurdaca/work/aicoe/storages/thoth/storages/sync.py", line 131, in sync_solver_documents
    graph.sync_solver_result(document, force=force)
  File "/home/fmurdaca/work/aicoe/storages/thoth/storages/graph/postgres.py", line 6264, in sync_solver_result
    solved, _ = Solved.get_or_create(
  File "/home/fmurdaca/.local/share/virtualenvs/storages-kD6ztVxe/lib64/python3.8/site-packages/sqlalchemy/orm/session.py", line 606, in __exit__
    self._assert_active(deactive_ok=True, prepared_ok=True)
  File "/home/fmurdaca/.local/share/virtualenvs/storages-kD6ztVxe/lib64/python3.8/site-packages/sqlalchemy/orm/session.py", line 305, in _assert_active
    raise sa_exc.ResourceClosedError(closed_msg)
sqlalchemy.exc.ResourceClosedError: This transaction is closed
```

pacospace (approved):
lgtm! 💯 Thanks!

PR #2604, opened 2022-02-28
added ordering to python versions endpoint

Gkrumbach07:
https://github.com/thoth-station/user-api/issues/1670

- [X] No

The new param is optional so older uses will work the same unless using the new param.

- [X] Yes

Added a new param to `get_solved_python_package_versions_all` called `order_by`. It optionally takes either `ASC` or `DESC` as input which orders the output by the version number.

## Description

It specifically will apply the postgresql query `ORDER BY string_to_array(regexp_replace(package_version, '[^0-9.]+|(?<=\d{4})(\d+)', '', 'g'), '.', '')::int[]`:
- regex replace function `regexp_replace(package_version, '[^0-9.]+|(?<=\d{4})(\d+)', '', 'g')` replaces all non numbers exclduing decimals with "", it also replaces any numbers that have more than 4 digits (to prevent integer overflow) ex) `1234567890 -> 1234`
     - ex) `"1.2.0rc12243432134"` -> `"1.2.01224"`
-  `string_to_array('1.2.01224', '.', '')` will split the string by delimiter=`.`, where empty strings are converted to NULL. ex)`"1.2."` -> `["1", "2", NULL]`
     - ex) `1.2.01224` -> `["1", "2", "1224"]`
- `::int[]` will cast the strings to numbers
- `ASC` | `DESC` will order by ascending or decending

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -1086,9 +1086,17 @@ class GraphDatabase(SQLBase):
             )
 
             if order_by == "ASC":
-                query = query.order_by(text("string_to_array(regexp_replace(package_version, '[^0-9.]+|(?<=\d{4})(\d+)', '', 'g'), '.', '')::int[] ASC"))
+                query = query.order_by(
+                    text(
+                        "string_to_array(regexp_replace(package_version, '[^0-9.]+|(?<=\d{4})(\d+)', '', 'g'), '.', '')::int[] ASC"
+                    )
+                )
             elif order_by == "DESC":
-                query = query.order_by(text("string_to_array(regexp_replace(package_version, '[^0-9.]+|(?<=\d{4})(\d+)', '', 'g'), '.', '')::int[] DESC"))
+                query = query.order_by(
+                    text(
+                        "string_to_array(regexp_replace(package_version, '[^0-9.]+|(?<=\d{4})(\d+)', '', 'g'), '.', '')::int[] DESC"
+                    )
+                )
 
             query = query.offset(start_offset).limit(count)
 
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -45,6 +45,7 @@ from packaging.version import parse as parse_version
 from sqlalchemy import create_engine
 from sqlalchemy import desc
 from sqlalchemy import func
+from sqlalchemy import text
 from sqlalchemy import exists
 from sqlalchemy import and_
 from sqlalchemy import tuple_
@@ -1055,6 +1056,7 @@ class GraphDatabase(SQLBase):
         *,
         start_offset: int = 0,
         count: Optional[int] = DEFAULT_COUNT,
+        order_by: Optional[str] = None,
         os_name: Optional[str] = None,
         os_version: Optional[str] = None,
         python_version: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -1083,6 +1085,19 @@ def get_solved_python_package_versions_all(
                 is_missing=is_missing,
             )
 
+            if order_by == "ASC":
+                query = query.order_by(
+                    text(
+                        "string_to_array(regexp_replace(package_version, '[^0-9.]+|(?<=\d{4})(\d+)', '', 'g'), '.', '')::int[] ASC"
```
fridex, line 1091: I'm very unsure about this ordering. There is a [specification](https://peps.python.org/pep-0440/) that has its semantics. That's why there is an implementation in `packaging` that is respecting the specification. What is the use case here? Do you want to show packages sorted on UI?

Gkrumbach07: The endpoint does not always return a sorted array. As described in this [issue](https://github.com/thoth-station/jupyterlab-requirements/issues/620#issuecomment-1043160872) with boto3. The reason for the odd ordering is because some (very few) versions in the DB are not formatted correctly which would cause errors if not cleaned first.

Do you think the sorting is unnecessary or is it being sorted incorrectly?

fridex: The ordering could be done in user-api. The implementation would require thoth-python and sort based on the semantic version - we have already this implementation so we could plug it there.

The main concern I have - pagination could easily break ordering easily (as it needs to be done outside of the database). It might be there are only a few versions per package, so it could be ok to drop pagination on these endpoint. wdyt?

Gkrumbach07: The reason I opted to preform the sorting on the database end is because it could come before pagination. Otherwise, like you said, sorting might on sort the defined pagination range and not the entire set as a whole. I found that to be a more efficient use of resources while still keeping pagination in place.

However I did think of sorting on the user-api due to its simplicity, however it could make some (not all) requests slow due to having to respond with the entire list vs a page of it. And that speed difference might be negligible for our purposes anyway.

We could always grab all the versions -> sort them -> and then paginated them all on the user api. Its probably the most brute force method but might not be that bad considering the small set sizes.

Let me know what you think and I can make the proper changes to the PRs

fridex: > The reason I opted to preform the sorting on the database end is because it could come before pagination. Otherwise, like you said, sorting might on sort the defined pagination range and not the entire set as a whole. I found that to be a more efficient use of resources while still keeping pagination in place.
> 
> However I did think of sorting on the user-api due to its simplicity, however it could make some (not all) requests slow due to having to respond with the entire list vs a page of it. And that speed difference might be negligible for our purposes anyway.
> 
> We could always grab all the versions -> sort them -> and then paginated them all on the user api. Its probably the most brute force method but might not be that bad considering the small set sizes.
> 
> Let me know what you think and I can make the proper changes to the PRs

As discussed on the last sig call - let's obtain all the versions we have in the database and sort them on user-api side. We have methods [for sorting already implemented in the thoth-python module](https://github.com/thoth-station/python/blob/b445cd1f81dc52b4f2b16284a160277b8833209e/thoth/python/package_version.py#L61).

The assumption here is that we have just a few versions per package (tens, rarely hundreds). This sorting should not take much time. Also, the sorting should be done in user-api code and not in thoth-storages module to avoid dependency hell.

fridex: With this change, pagination for package versions can be dropped completely on the user-api endpoint.

Gkrumbach07: Will close this PR then as storages does not need to be changed

PR #2621, opened 2022-04-11
Increase default S3 timeout to 180s for retrieving data from Ceph

mayaCostantini:

- No

- Yes

Retrieving a large list of objects from an S3 bucket can cause a gateway timeout error because the default `botocore` client configuration sets the timeout to 60 seconds. Increasing the timeout to 180 seconds will allow for more objects to be retrieved without a failure. As an example, getting a hundred Amun inspections from the `opf-datacatalog` public bucket takes approximately 2 minutes, which exceeds the default timeout value.

### DIFF
--- a/thoth/storages/ceph.py
+++ b/thoth/storages/ceph.py
@@ -146,7 +146,7 @@ class CephStore(StorageBase):
         # signature version is needed to connect to new regions which
         # support only v4
         self._s3 = session.resource(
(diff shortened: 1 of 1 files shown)


schwesig (approved):
/lgtm
/approve

fridex (approved):
Oh, very nice! 💯 👍🏻 
Thanks!

PR #2625, opened 2022-04-12
Randomize solver document listing

mayaCostantini:

Related to https://github.com/thoth-station/thoth-application/issues/2513
- [ ] make sure the [sync method in thoth-storages ](https://github.com/thoth-station/storages/blob/d58f8a72787aebcc20a2ff702dc087d1286384dc/thoth/storages/sync.py#L117)can randomize document listing so that we can run multiple sync jobs in parallel

- No

Randomizing the solver documents listing is optional.

- Yes

Introduce an optional argument to the [`sync_solver_documents`](https://github.com/thoth-station/storages/blob/d58f8a72787aebcc20a2ff702dc087d1286384dc/thoth/storages/sync.py#L95) method in storages to randomize solver document listing.

### DIFF
--- a/thoth/storages/sync.py
+++ b/thoth/storages/sync.py
@@ -39,6 +39,7 @@ from .solvers import SolverResultsStore
 from .graph import GraphDatabase
 
 _LOGGER = logging.getLogger(__name__)
+RANDOMIZE_LISTING = os.getenv("RANDOMIZE_LISTING", 0)
 
 
 def sync_adviser_documents(
@@ -99,11 +100,8 @@ def sync_solver_documents(
     graceful: bool = False,
     graph: Optional[GraphDatabase] = None,
     is_local: bool = False,
-    randomize_listing: Optional[bool] = False,
 ) -> tuple:
     """Sync solver documents into graph."""
-    document_listing = list(solver_store.get_document_listing())
-
     if is_local and not document_ids:
         raise ValueError(
             "Cannot sync documents from local directory without explicitly specifying a list of documents to be synced"
@@ -117,10 +115,13 @@ def sync_solver_documents(
         solver_store = SolverResultsStore()
         solver_store.connect()
 
-    if randomize_listing:
+    if not document_ids:
+        document_listing = list(solver_store.get_document_listing())
+
+    if RANDOMIZE_LISTING:
         if document_ids:
             random.shuffle(document_ids)
-        elif document_listing:
+        else:
             random.shuffle(document_listing)
 
     processed, synced, skipped, failed = 0, 0, 0, 0
--- a/thoth/storages/sync.py
+++ b/thoth/storages/sync.py
@@ -25,6 +25,7 @@ from typing import Tuple
 from typing import List
 from typing import Optional
 from pathlib import Path
+from random import random
 
 from .analyses import AnalysisResultsStore
 from .advisers import AdvisersResultsStore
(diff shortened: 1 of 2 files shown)


thoth/storages/sync.py
```diff
@@ -113,8 +117,14 @@ def sync_solver_documents(
...
 
+    if randomize_listing:
+        if document_ids:
+            random.shuffle(document_ids)
+        elif document_listing:
+            random.shuffle(document_listing)
+
     processed, synced, skipped, failed = 0, 0, 0, 0
-    for document_id in document_ids or solver_store.get_document_listing():
+    for document_id in document_ids or document_listing:
```
fridex, line 127: Please shuffle both to be consistent.

thoth/storages/sync.py
```diff
@@ -98,8 +99,11 @@ def sync_solver_documents(
     graceful: bool = False,
     graph: Optional[GraphDatabase] = None,
     is_local: bool = False,
+    randomize_listing: Optional[bool] = False,
 ) -> tuple:
     """Sync solver documents into graph."""
+    document_listing = list(solver_store.get_document_listing())
```
fridex, line 105: This looks like an unnecessary step if `document_ids` is provided.

thoth/storages/sync.py
```diff
@@ -98,8 +99,11 @@ def sync_solver_documents(
     graceful: bool = False,
     graph: Optional[GraphDatabase] = None,
     is_local: bool = False,
+    randomize_listing: Optional[bool] = False,
```
fridex, line 102: Can we have this configurable via environment variables on a library level? That way we can control this behavior directly from deployment manifests.

mayaCostantini:
Thanks for the review @fridex 💯 Should we also randomize the listing in other sync methods?

fridex:
> Thanks for the review @fridex 100 Should we also randomize the listing in other sync methods?

+1, that would be great

Gregory-Pereira:
Will we need to write a separate script that will do this conversion /migration for everything currently in the DB?

Gregory-Pereira:
Also shouldn't [this micro version](https://github.com/thoth-station/storages/pull/2625/files#diff-b384062aa4e5e15a814b9992d5362f27bcda6f8a3e016f8095d8bc7e4a5a96e1L55) be bumped with this feature.

Gregory-Pereira:
Also shouldn't [this micro version](https://github.com/thoth-station/storages/pull/2625/files#diff-b384062aa4e5e15a814b9992d5362f27bcda6f8a3e016f8095d8bc7e4a5a96e1L55) be bumped with this feature.

thoth/storages/sync.py
```diff
@@ -61,8 +63,17 @@ def sync_adviser_documents(
...
+
+    if RANDOMIZE_LISTING:
+        if document_ids:
+            random.shuffle(document_ids)
+        else:
+            random.shuffle(document_listing)
+
     processed, synced, skipped, failed = 0, 0, 0, 0
-    for document_id in document_ids or adviser_store.get_document_listing():
+    for document_id in document_ids or document_listing:
```
fridex, line 76: The snippet above might be prone to programmatic errors as the if-else is basically done twice. It might be a good idea to be explicit which listing is actually used just once:

```suggestion
    if _RANDOMIZE_LISTING:
        listing = random.shuffle(document_ids or document_listing)
    for document_id in listing:
```

thoth/storages/sync.py
```diff
@@ -61,8 +63,13 @@ def sync_adviser_documents(
         adviser_store = AdvisersResultsStore()
         adviser_store.connect()
 
+    listing = list(adviser_store.get_document_listing()) or document_ids
```
fridex, line 66: Can we cast the generator to a list only if randomization is on? I know it might be nitpicking, but in some cases it might improve perf :)

mayaCostantini: No you're right, thanks! :100: Added in https://github.com/thoth-station/storages/pull/2625/commits/03acd31c136b92423ef836ac38869fed6fd5e13c

fridex (approved):
lgtm, thanks! 💯 👍🏻

PR #2639, opened 2022-05-23
added option for starts_with filter on /python/package

Gkrumbach07:
https://github.com/thoth-station/user-api/issues/1743

- [X] No

- [X] Yes

Adds an option to query the `get_python_package_version_names_all` function with a new param `starts_with`
## Description
starts_with will filter the query to only include packages that have a name that starts with starts_with input value


mayaCostantini (approved):
lgtm :+1: :100:

PR #2655, opened 2022-07-08
Fix flake8 errors and enable pre-commit hook

VannTen:
- Factor out flake8 config from pre-commit
- Fix flake8 errors and enable pre-commit hook

…

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Note for reviewers

Particular things I'm not sure about :

```
diff --git a/thoth/storages/graph/postgres.py b/thoth/storages/graph/postgres.py
index 178571a..d887523 100644
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -6517,7 +6517,7 @@ class GraphDatabase(SQLBase):
                 # Discover unresolved package matching the message ID
                 if package["id"] == "MISSING-PACKAGE":
 
-                    python_package_version_entity = self._create_python_package_version(
+                    self._create_python_package_version(
                         session,
                         package_name=package["package_name"],
                         package_version=package["package_version"].lstrip("=="),
@@ -6888,10 +6888,13 @@ class GraphDatabase(SQLBase):
```
Since the variable was not used, I suppose the purpose is only to persist the
data ?

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -375,7 +375,7 @@ class GraphDatabase(SQLBase):
 
     def get_table_alembic_version_head(self) -> str:
         """Get alembic version head from database table."""
-        query = f"SELECT * FROM alembic_version"
+        query = "SELECT * FROM alembic_version"
 
         with self._session_scope() as session:
             result = session.execute(query).fetchone()
@@ -400,7 +400,7 @@ class GraphDatabase(SQLBase):
 
     def get_alembic_version_count_all(self) -> int:
         """Get number of records in alembic version table (1 expected)."""
-        query = f"SELECT COUNT(*) FROM alembic_version"
+        query = "SELECT COUNT(*) FROM alembic_version"
 
         with self._session_scope() as session:
             result = session.execute(query).fetchone()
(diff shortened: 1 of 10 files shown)


thoth/storages/graph/postgres.py
```diff
@@ -6517,7 +6517,7 @@ def sync_provenance_checker_result(self, document: dict) -> None:
                 # Discover unresolved package matching the message ID
                 if package["id"] == "MISSING-PACKAGE":
 
-                    python_package_version_entity = self._create_python_package_version(
+                    self._create_python_package_version(
```
mayaCostantini, line 6520: I agree on keeping the unused variable `provenance_checker_run` in this method as this looks like an optimization to persist the data within the session scope :+1: This has been used in other methods where `get_or_create` was called as well in this class.

mayaCostantini (approved):
lgtm, thanks! :+1: :100:

thoth/storages/graph/postgres.py
VannTen: In fact, I did not change that line (provenance_checker_run) because it was not flagged by flake8. It's flagged by pylint though apparently, not sure why there is a discrepancy...

PR #2661, opened 2022-07-19
WIP: Provide high-level overview of Thoth storage model

### LINKED ISSUE #2658: Developper documentation relation postgres <--> Ceph
It's not very clear from the code or the documentation what Ceph is used for:

I can see [here](https://thoth-station.ninja/docs/developers/storages/index.html#syncing-results-of-a-workflow-run-in-the-cluster) that we store task results,
[here](https://thoth-station.ninja/docs/developers/adviser/developers_guide.html?highlight=ceph#testing-application-against-ceph-and-a-knowledge-graph-database)
how to access it, but there should a high level description of it's role,
comparable to the postgres schema that we can generate with `generate-schema`.

Currently it's a bit hard to get a definite idea of what is inside (and
consequently, what should be deleted, see #2657)

Also, is it only used as an S3 store ? In that case, why reference Ceph
particularly ?

/kind documentation

VannTen:

closes #2658
…

## Description

This intend to improve the thoth-storages modules docs to give a high-level
overview of how the data is organized, and differents items relates together.

The intended public would be :

- Thoth developers, who need to understand the storage model in order to work on
  the storage implementation
- Thoth operators, to understand which data is valuable and the relations
  between data items: what can be recreated, at what cost, etc. + what must be
  backed up atomically (avoid inconsistency)
 
Ideally, it should stays generic but points to part of the code / generated
documentation which can list all the concerned objects and in what "class" they
fall.

# Current state

This is intended as a continuation of the conversation in #2658, so the
information might not be accurate yet, it reflects my current comprehension /
hypotheses 

# Goals

- [ ] Developer documentation
- [ ] Operational documentation

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,6 +23,28 @@ This library provides a library called `thoth-storages
 for `PostgreSQL database <https://www.postgresql.org/>`__ as well as adapters
 for manipulating with `Ceph <https://ceph.io/>`__ via its S3 compatible API.
 
+Overview
+========
+
+Thoth stores its data in two persistent stores, a Postgres SQL database, and an
+S3 object store.
+
+S3 Object store
+---------------
+
+The S3 store contains JSON documents, which are the results of Thoth operations
+(analysis, reports, etc).
+
+Postgres Database
+-----------------
+
+Two kinds of things are stored in the database:
+
+- The aforementioned results documents are referenced.
+- Reference to packages sources (ex: Python packages indexes).
+
+These two combined form the knowledge graph.
+
 Quick Start
 ===========
 


VannTen:
As noted in the linked issue, maybe operationnal knowledge should **not** be
part of storages, but in a central places regardless of the part of Thoth.

So the current diff (only dev / internal storage logic) will be a start.

README.rst
```diff
@@ -23,6 +23,28 @@ This library provides a library called `thoth-storages
...
+
+Postgres Database
+-----------------
+
+Two kinds of things are stored in the database:
+
+- The aforementioned results documents are referenced.
+- Reference to packages sources (ex: Python packages indexes).
+
+These two combined form the knowledge graph.
```
mayaCostantini, line 46: I don't know if we can summarize the content of Thoth's postgres database in such a simple way. In fact, the DB contains all kind of information not directly related to document results and package indexes, see the [database model](https://github.com/thoth-station/storages/blob/master/thoth/storages/graph/models.py) for an overview of the content available.

VannTen:
> I don't know if we can summarize the content of Thoth's postgres database in such a simple way. In fact, the DB contains all kind of information not directly related to document results and package indexes, see the [database model](https://github.com/thoth-station/storages/blob/master/thoth/storages/graph/models.py) for an overview of the content available.

Can't we classify it somehow ? I had taken a (actually, some ^) look at
the models class before, and I can see there is some common patterns
between the differents models, but I'm not sure what those patterns are.

Maybe we should have more parents class regrouping related models, which
would also serve as self-documenting the code ? (by being "category" of
things stored in the DB)

mayaCostantini:
> I can see there is some common patterns between the differents models, but I'm not sure what those patterns are. Maybe we should have more parents class regrouping related models, which would also serve as self-documenting the code ? (by being "category" of things stored in the DB)

I don't think it would be a good idea to group models under more generic objects, as declaring one class per table seems like a clearer implementation wrt SQLAlchemy ORM conventions. Do you have any examples of classes you think should be more generic to help me understand your point?

VannTen:
Sure. For example, SoftwareEnvironment and ExternalSoftwareEnvironment 

> I don't think it would be a good idea to group models under more
> generic objects, as declaring one class per table seems like a clearer
> implementation wrt SQLAlchemy ORM conventions.

Theirs docs at least suggest that inheritance is a valid way to work with the
ORM (using either [joined table or single table
inheritance](https://docs.sqlalchemy.org/en/14/orm/inheritance.html))

Issue #2673, opened 2022-08-08
Enable mypy hook


VannTen:
## Problem statement
We have problems like #2671 (functions got modified, call site was not)
If we have one, we likely have more. 

Mypy would catch that sort of stuff. We need to enable the hook and fix
whatever mistakes there is (141 at the moment, with a bunch of dupli/triplicates
and more)

## Acceptance Criteria
- [ ] mypy hooks is enabled

/assign
/priority important-soon

PR #2674, opened 2022-08-08
Enable mypy pre-commit hook

VannTen:
see #2673


mayaCostantini (approved):
lgtm, thanks! :100

VannTen:
I only enabled the hook, still needs to gruntily fix all the errors ^

VannTen:
Related to the errors in `thoth/storages/graph/models_base.py`:
https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html

TL;DR (from my understanding)typing support is not super practical in SQLAlchemy
declarative API in 1.3/1.4. Either needs a mypy plugin (which is alpha and
deprecated for sqlalchemy 2.0) or switching to 2.0 (in dev).

VannTen:
Note: in 2e5cd41 I explicited what I assumed to be the intent. For review I'll
need to conform that assumption

VannTen:
The `graph/*` errors are a bit hard to fix without restructuring quite intensely
the code, so I think we could benefits from merging this as is. (Finding a way
to exclude thoth.storages.graph module would be great, but for now I don't know
how (see https://github.com/python/mypy/issues/10820))

For reviewers: I think most of the commits are reviewable individually. In fact
they can probably be cherry-picked individually as well.


Some are "breaking" changes to the API, (in the types) but in most cases the
underlying implementation was already breaking stuff (or the call sites used the
code correctly despite wrong types ! ^).

I will detail the commits changing the API in a later comment.

VannTen:
Commits breaking API.

- 5f9faf4 Fix incompatible method overload for AnalysisByDigest
  Only call site has a pending PR
- solvers:
  bcc355c Fix incompatible child class method override
  369d2e2 Use TypedDict for reifying "all or nothing" for solver infos
  -> Align the API + use types to states parameters requirements -> api change 
- 2c6e75c sync: More precise handlers types
  Not breaking (List are Iterable)
  dd00f78 sync: remove unused parameters and refactor
  Align interface with behavior (removed parameters were not used)
- 2055ceb CephCache
  Correct return type
- 334ec44 dependency_monkeys
  Correct return type
- 24f637e Trying to fix get_document_ on solvers
    This avoid the needs for override on get_document_count but kinda drop type safety for it.
    Best option would be to make it a static method / utils and feed it the result of get_document_listing instead of including it in the class (since we don't do anything more efficient like using list_objects or something like that)

Issue #2666, opened 2022-07-25
Implement a method to retrieve TTL of cached result


mayaCostantini:
## Problem statement

Needed for https://github.com/thoth-station/user-api/issues/1614

Implement a method (for example in [ResultStorageBase](https://github.com/thoth-station/storages/blob/c70b89f809cdc67574617d3091f66231a6902e47/thoth/storages/result_base.py#L32)) to retrieve the Time To Live of a cached result based on the corresponding document id.

PR #2676, opened 2022-08-09
Retrieve TTL of cached documents

mayaCostantini:

Related to https://github.com/thoth-station/storages/issues/2666

- No

- Yes

Implement a method to retrieve the time to live of a cached document, considering S3 bucket timezones for different environments.
This implementation is based on the current documents caching time of 2 hours.

### DIFF
--- a/thoth/storages/ceph.py
+++ b/thoth/storages/ceph.py
@@ -101,12 +101,7 @@ class CephStore(StorageBase):
 
     def retrieve_document_last_modification_date(self, object_key: str) -> datetime:
         """Retrieve the last modification date of a document from S3."""
-        try:
-            return self._s3.Object(self.bucket, f"{self.prefix}{object_key}").get()["LastModified"]
-        except botocore.exceptions.ClientError as exc:
-            if exc.response["Error"]["Code"] in ("404", "NoSuchKey"):
-                raise NotFoundError("Failed to retrieve object, object {!r} does not exist".format(object_key)) from exc
-            raise
+        return self._s3.Object(self.bucket, f"{self.prefix}{object_key}").get()["LastModified"]
 
     def iterate_results(self, prefix_addition: typing.Optional[str] = None) -> typing.Generator[tuple, None, None]:
         """Iterate over results available in the Ceph."""
--- a/thoth/storages/ceph.py
+++ b/thoth/storages/ceph.py
@@ -24,8 +24,6 @@ import typing
 import boto3
 import botocore
 
-from datetime import datetime
-
 from .base import StorageBase
 from .exceptions import NotFoundError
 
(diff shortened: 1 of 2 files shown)


thoth/storages/ceph.py
```diff
@@ -97,6 +99,15 @@ def retrieve_blob(self, object_key: str) -> bytes:
...
             raise
 
+    def retrieve_document_last_modification_date(self, object_key: str) -> datetime:
+        """Retrieve the last modification date of a document from S3."""
+        try:
+            return self._s3.Object(self.bucket, f"{self.prefix}{object_key}").get()["LastModified"]
+        except botocore.exceptions.ClientError as exc:
+            if exc.response["Error"]["Code"] in ("404", "NoSuchKey"):
+                raise NotFoundError("Failed to retrieve object, object {!r} does not exist".format(object_key)) from exc
+            raise
```
VannTen, line 109: Do we really need to catch exceptions at that level ?

Since we're re-raising anyway (-> so the caller has to handle it itself), I would drop that.

thoth/storages/ceph.py
```diff
@@ -97,6 +99,15 @@ def retrieve_blob(self, object_key: str) -> bytes:
                 raise NotFoundError("Failed to retrieve object, object {!r} does not exist".format(object_key)) from exc
             raise
 
+    def retrieve_document_last_modification_date(self, object_key: str) -> datetime:
+        """Retrieve the last modification date of a document from S3."""
+        try:
+            return self._s3.Object(self.bucket, f"{self.prefix}{object_key}").get()["LastModified"]
```
VannTen, line 105: Since the previous comment would make this a single line, maybe we could just drop the method and have the caller use `['LastModified'] on the object` ?

thoth/storages/ceph_cache.py
```diff
@@ -36,3 +37,17 @@ def retrieve_document_record(self, document_id: str) -> dict:
...
         """Store the given document record in the cache."""
         self.ceph.store_document(document, document_id)
+
+    def retrieve_document_ttl(self, document_id: str) -> float:
+        """Retrieve the TTL for a cached document."""
+        # Verify if the document is present in the cache
+        self.retrieve_document_record(document_id)
+
+        # No timezone is indicated to calculate the present datetime on purpose to get the correct time delta
+        # regardless of the environment where the data is retrieved from (production or stage)
```
VannTen, line 47:
```suggestion
        # Uses UTC time to be environment agnostic (no timezone)
```
I don't think we should point out specifics environments, thoth could be deployed on others.

VannTen (changes requested):
If I'm going against the codebases convention here. feel free to point out :+1:

thoth/storages/ceph.py
```diff
@@ -97,6 +99,10 @@ def retrieve_blob(self, object_key: str) -> bytes:
                 raise NotFoundError("Failed to retrieve object, object {!r} does not exist".format(object_key)) from exc
             raise
 
+    def retrieve_document_last_modification_date(self, object_key: str) -> datetime:
+        """Retrieve the last modification date of a document from S3."""
+        return self._s3.Object(self.bucket, f"{self.prefix}{object_key}").get()["LastModified"]
```
VannTen, line 104: I feel like we should probably not have a method for something so specific.

What about 
```suggestion
    def retrieve_document_attr(self, object_key: str, attr: str) -> datetime:
        """Retrieve the given attribute of a document from S3."""
        return self._s3.Object(self.bucket, f"{self.prefix}{object_key}").get()[str]
```
and calling that with "LastModified" ?

Maybe I'm worried over nothing. I'm not sure of the correct ratio specialize/generalize.

VannTen: (leaving the method out completely like I initially suggested does not seem practical)

mayaCostantini:
@VannTen is this ready to be merged?

VannTen:
I think we could generalize `retrieve_document_last_modification_date`
as mentioned in the review, otherwise, the rest looks OK to me.

mayaCostantini:
@VannTen thanks for the reviews, changes added :+1:

VannTen (approved):

PR #2677, opened 2022-08-10
Make _iter_dates_prefix_addition a static method

### LINKED ISSUE #2671: Bad usage of _iter_dates_prefix in solver.py
## Bug description
In thoth/storages/solver.py, `_iter_dates_prefix_addition` is used with
`prefix`, a parameter which it does not have anymore.
This might not cause a bug yet, but it certainly will.

### Additional context

This was caused by #2609

### Thoughts

We need to enable mypy to catch that sort of stuff for us...

/kind bug
/assign

VannTen:
see commit message. Fix #2671

### DIFF
--- a/thoth/storages/ceph.py
+++ b/thoth/storages/ceph.py
@@ -57,9 +57,9 @@ class CephStore(StorageBase):
         if not self.prefix.endswith("/"):
             self.prefix += "/"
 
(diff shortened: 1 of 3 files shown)


mayaCostantini (approved):
lgtm, thanks for the fix :100:

PR #2702, opened 2022-09-14
Convert ceph datetime object to be offset-aware

mayaCostantini:

Related to https://github.com/thoth-station/user-api/issues/1614
Fix required for https://github.com/thoth-station/user-api/pull/1798

- No

- Yes

Convert `self.ceph.retrieve_document_attr(object_key=document_id, attr="LastModified")` object to an offset-aware datetime object to fix the diff between `datetime.now()` (offset-aware) and `self.ceph.retrieve_document_attr(object_key=document_id, attr="LastModified")` (offset-naive).

### DIFF
--- a/thoth/storages/ceph_cache.py
+++ b/thoth/storages/ceph_cache.py
@@ -45,7 +45,10 @@ class CephCache(ResultStorageBase):
 
         # Uses UTC time to be environment agnostic (no timezone)
         time_lived = (
-            datetime.now() - self.ceph.retrieve_document_attr(object_key=document_id, attr="LastModified")
+            datetime.now()
+            - self.ceph.retrieve_document_attr(object_key=document_id, attr="LastModified").replace(
+                tzinfo=datetime.timezone.utc
+            )
         ).total_seconds()
 
         if time_lived <= 14400.0:


thoth/storages/ceph_cache.py
```diff
@@ -45,7 +45,10 @@ def retrieve_document_ttl(self, document_id: str) -> float:
 
         # Uses UTC time to be environment agnostic (no timezone)
         time_lived = (
-            datetime.now() - self.ceph.retrieve_document_attr(object_key=document_id, attr="LastModified")
+            datetime.now()
+            - self.ceph.retrieve_document_attr(object_key=document_id, attr="LastModified").replace(
+                tzinfo=datetime.timezone.utc
```
harshad16, line 50: str.replace format are normal in `,` format right?
should this be:
```suggestion
                tzinfo, datetime.timezone.utc
```

harshad16 (changes requested):
just one doubt

thoth/storages/ceph_cache.py
harshad16: Thanks for the clarification.

harshad16 (approved):
/lgtm
:+1:

Issue #2703, opened 2022-09-14
package_license metadata throwing attribute error


harshad16:
## Bug description

The graph sync job which uses storage to sync data to the database is failing.
with the following issue:
```
  File "app.py", line 125, in _do_sync
    stats = sync_documents(
  File "/opt/app-root/lib64/python3.8/site-packages/thoth/storages/sync.py", line 616, in sync_documents
    stats_change = handler(
  File "/opt/app-root/lib64/python3.8/site-packages/thoth/storages/sync.py", line 143, in sync_solver_documents
    graph.sync_solver_result(document, force=force)
  File "/opt/app-root/lib64/python3.8/site-packages/thoth/storages/graph/postgres.py", line 6051, in sync_solver_result
    license_name=package_license["license"].get("full_name"),
AttributeError: 'str' object has no attribute 'get'
```

### Steps to Reproduce
1. Schedule an solver run
2. See error

### Additional context
[solver-rhel-8-py38-220914201834-2e90a53ac16cc0b0-3334486321-main.log](https://github.com/thoth-station/storages/files/9570352/solver-rhel-8-py38-220914201834-2e90a53ac16cc0b0-3334486321-main.log)

harshad16:
/sig stack-guidance
/priority critical-urgent

This is required for unblocking ingestion.

mayaCostantini:
From the logs, it seems like we don't handle properly the case when the license is undetected:

```
license {'license': 'UNDETECTED', 'license_identifier': 'UNDETECTED', 'license_version': 'UNDETECTED', 'warning': True}
```

We should catch the `UNDETECTED` values and set the corresponding fields to None.

harshad16:
This has been fixed with help of the pr and issue linked in this issue.
with update in database with https://github.com/thoth-station/storages/issues/2713#issuecomment-1252833725
we can be close this for now.
thanks everyone for working on this.

PR #2705, opened 2022-09-15
Fix package_license metadata attribute error

mayaCostantini:

Fixes https://github.com/thoth-station/storages/issues/2703

- No

- Yes

Handle the case when the license metadata is undetected during document sync job.

### DIFF
--- a/thoth/storages/graph/postgres.py
+++ b/thoth/storages/graph/postgres.py
@@ -6046,11 +6046,25 @@ class GraphDatabase(SQLBase):
                         "found in PythonPackageMetadata table, the error is not fatal"
                     )
 
+                if package_license["license"] == "UNDETECTED":
+                    package_license_name = None
+                    package_license_spdx = None
+
+                else:
+                    package_license_name = package_license["license"].get("full_name")
+                    package_license_spdx = package_license["license"].get("identifier_spdx")
+
+                if package_license["license_version"] == "UNDETECTED":
+                    package_license_version = None
+
+                else:
+                    package_license_version = package_license["license_version"]
+
                 license_metadata, _ = PythonPackageLicense.get_or_create(
                     session,
-                    license_name=package_license["license"].get("full_name"),
-                    license_identifier=package_license["license"].get("identifier_spdx"),
-                    license_version=package_license["license_version"],
+                    license_name=package_license_name,
+                    license_identifier=package_license_spdx,
+                    license_version=package_license_version,
                 )
 
                 try:


VannTen (approved):
Do we have some kind of schema for the solver document ? If I read the code right, `license` is either a dict or a string, right ?

(Thinking forward for future stricter typing in the graph module)

In any case,
/lgtm

harshad16:
From the look of the license-solver the license bit should be dict
https://github.com/thoth-station/license-solver/blob/28c29dbf560c7165fd67419a3abfe3b53a1c6f1a/thoth/license_solver/package.py#L57
there was issue in default value sent in case of uncatched license.
There is already a pr for fix.
https://github.com/thoth-station/license-solver/pull/67

I think that would fix the issue, and we would have to catch this in storages.

harshad16:
> Hum. Reading that code, maybe `license` should be an `Optional[Dict[str, str]]` and replace UNDETECTED with None ? Unless something else can be there in place of UNDETECTED ?

maybe it , seems like the code of license solver needs more of a closer look.
As of now, we can follow the pattern of existing one i feel
and explore more going forward.

in any case, i feel we might not have to do the type checking ?
or do you feel it just better to check as there is no much effect on runtime?

VannTen:
On Thu, Sep 15, 2022 at 08:11:24AM -0700, Harshad Reddy Nalla wrote:
> in any case, i feel we might not have to do the type checking ?
> or do you feel it just better to check as there is no much effect on runtime?

I'm not the person to ask ^^. I'm totally sold on strong static
type-checking, so I'll always feels more type-correctness is good to
have.
That's a time allocation problem, though (for example the whole graph
module in storages is a bit much) ; so if we solve the problem at hand,
that could probably wait.

> no much effect on runtime

Type hints are discarded at runtime (at parsing time I guess) so makes
that virtually no effect.


---
Don't know why github keeps dropping my email replies :/

thoth/storages/graph/postgres.py
```diff
@@ -6046,11 +6046,25 @@ def sync_solver_result(self, document: dict, *, force: bool = False) -> None:
...
+                    package_license_version = package_license["license_version"]
+
                 license_metadata, _ = PythonPackageLicense.get_or_create(
                     session,
-                    license_name=package_license["license"].get("full_name"),
-                    license_identifier=package_license["license"].get("identifier_spdx"),
-                    license_version=package_license["license_version"],
+                    license_name=package_license_name,
+                    license_identifier=package_license_spdx,
+                    license_version=package_license_version,
```
harshad16, line 6067: These field are non-nullable based on the scehma 
https://github.com/thoth-station/storages/blob/acd96e535a61e6f6d5d9ac857c38eef6de4fdaa8/thoth/storages/graph/models.py#L106

we should raise exception if the package_license info is not provided.

harshad16 (changes requested):
based on further investigation
if we want to do more type checking 
we should raise exceptions instead of placing null values.

thoth/storages/graph/postgres.py
VannTen: I just took a look at spdx : https://scancode-licensedb.aboutcode.org/index.html.
Instead of None, I think we should refer to the spdx classification. See : 
- [no-license](https://scancode-licensedb.aboutcode.org/no-license.html)
- [unknown](https://scancode-licensedb.aboutcode.org/unknown.html)
- [unknown-spdx](https://scancode-licensedb.aboutcode.org/unknown-spdx.html)

Might be more.

mayaCostantini: Is there any particular reason `python_package_license` are not nullable? Could we change the schema to allow them to be nullable, or would this require too many changes in license-solver code?

VannTen: I haven't investigated, but I'm not sure we should, given the alternatives of the mentioned "pseudo" spdx licences.

First, it makes us use an established standard.
Second, nullable types are always trickier to handle correctly (because that mean special casing unless you use a level of type-machinery I don't think Python)

harshad16:
This might need more static type-checking.
for now, the already include changes are keeping this in place.
we should work on license solver to get the default value and then include the type-checking.
Closing this for now.
lets address this in future, if we face same issue.
thanks for all the discussions and work.

PR #2709, opened 2022-09-19
Set license solver params as unique constraint to reduce duplication

harshad16:
Set license solver params as unique constraint to reduce duplication

Related-to: https://github.com/thoth-station/storages/issues/2703

## Description

This would fix the schema.


mayaCostantini (approved):
lgtm, thanks for the fix :100:

thoth/storages/data/alembic/versions/7f7411e247d6_migrate_license_solver_data.py
```diff
@@ -0,0 +1,30 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_unique_constraint(
+        None, "python_package_license", ["license_name", "license_identifier", "license_version"]
+    )
+    # ### end Alembic commands ###
```
VannTen, line 24: I have some concerns over this.
I don't think create_unique_constraint automatically find and eliminates duplicates; so the autogenerated migration will not be good enough, we'd need to:
1. Find first ocurrence for each case
2. Update foreign key in other table to refer to first ocurrence
3. Remove duplicates.
Otherwise we're going to run into an IntegrityError, I think.

thoth/storages/graph/models.py
```diff
@@ -109,6 +109,7 @@ class PythonPackageLicense(Base, BaseExtension):
 
     # Relations
     python_package_version = relationship("PythonPackageVersion", back_populates="licenses")
+    __table_args__ = (UniqueConstraint("license_name", "license_identifier", "license_version"),)
```
VannTen, line 112: Could we not have a unique constraint on license_identifier only ?
If I'm not mistaken, this allows duplicates license_identifier for example, as long as the two other fields are not identical.

Issue #2722, opened 2022-09-26
Write unit tests for Ceph cache interface


mayaCostantini:
## Problem statement

Unit test the following ceph_cache features:

- [ ] Retrieving documents from the cache
- [ ] Storing document records
- [ ] Retrieving document TTL

PR #2723, opened 2022-10-04
Unit test Ceph cache and fix timezone awareness

mayaCostantini:

Fixes https://github.com/thoth-station/storages/issues/2722
Related to https://github.com/thoth-station/storages/issues/2666

- No

- Yes

### DIFF
--- a/thoth/storages/ceph_cache.py
+++ b/thoth/storages/ceph_cache.py
@@ -46,11 +46,11 @@ class CephCache(ResultStorageBase):
 
         # Uses UTC time to be environment agnostic (no timezone)
         time_lived = (
-            datetime.now().replace(tzinfo=timezone.utc)
+            datetime.now(timezone.utc)
             - self.ceph.retrieve_document_attr(object_key=document_id, attr="LastModified").replace(tzinfo=timezone.utc)
         ).total_seconds()
 
         if time_lived <= 14400.0:
-            return time_lived
+            return 14400.0 - time_lived
 
         return 0.0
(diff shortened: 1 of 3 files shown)


thoth/storages/ceph_cache.py
```diff
@@ -46,11 +46,11 @@ def retrieve_document_ttl(self, document_id: str) -> float:
...
         # Uses UTC time to be environment agnostic (no timezone)
         time_lived = (
-            datetime.now()
+            datetime.now(timezone.utc)
             - self.ceph.retrieve_document_attr(object_key=document_id, attr="LastModified").replace(tzinfo=timezone.utc)
         ).total_seconds()
 
         if time_lived <= 14400.0:
-            return time_lived
+            return 14400.0 - time_lived
```
VannTen, line 54:
```suggestion
            return timedelta(hours=4) - time_lived
```
or if we don't want to return a timedelta 
```suggestion
            return (timedelta(hours=4) - time_lived).total_seconds()
```

thoth/storages/ceph_cache.py
```diff
@@ -46,11 +46,11 @@ def retrieve_document_ttl(self, document_id: str) -> float:
...
 
         return 0.0
```
VannTen, line 56:
```suggestion
        return timedelta()
```
Only if we switch to a timedelta return.

VannTen (changes requested):

VannTen:
Btw, side comment: I'm not sure `datetime` is a good fit for a TTL. For that we don't need fancy handling of timezone/DST, just a unix timestamp, don't we ?

`time` would probably better. Unfortunately s3 "lastmodified" only give us a datetime. Hum. Let's not bother for now, I'll keep that on a note and see if it warrant another issue.

Issue #2750, opened 2023-01-12
sync_document function is not able to traverse documents


harshad16:
## Bug description

The function [sync_document](https://github.com/thoth-station/storages/blob/master/thoth/storages/sync.py#L570) in storages/sync.py is unable to traverse the document ids. Causing issues like  `ValueError
No handler defined for document identifier /mnt/component/document/solver-rhel-8-py36-230111052532-e7dc96c907760295`
https://sentry.io/organizations/red-hat-it/issues/3867414521/?alert_rule_id=529086&alert_timestamp=1673413911988&alert_type=email&environment=ocp4-test&project=1298083

### Steps to Reproduce
1. Create a solver for ingestion
2. Wait for graph-sync workflow
3. Scroll down in log

### Expected behavior
Parse the document_id and sync the document to the appropriate handler.

### Environment information
`thoth-storages` in v0.73.6

PR #2751, opened 2023-01-12
:face_in_clouds: Fix the sync_documents formatting

harshad16:
Fix the sync_documents formatting

Fixes: https://github.com/thoth-station/storages/issues/2750

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

## Description

- Change `dict.fromkeys` to dict comprehension as dict.fromkeys behave weirdly when value is list.
More info: https://stackoverflow.com/questions/3000468/unwanted-behaviour-from-dict-fromkeys

- fix the handle assignment by changing the `doc.rfind` to `doc.find` , so it get correct prefix.
```
>>> x = "adviser-efa7213babd12911"
>>> x[x.rfind("/")+1: x.find("-")]
'adviser'
```

- we would only like to run sync handlers if the document is having value or None.

### DIFF
--- a/thoth/storages/sync.py
+++ b/thoth/storages/sync.py
@@ -580,12 +580,14 @@ def sync_documents(
     >>> from thoth.storages.sync import sync_documents
     >>> sync_documents(["adviser-efa7213babd12911", "package-extract-f8e354d9597a1203"])
     """
+
+    handlers: Dict[str, List[str]] = dict()
     if document_ids:
-        handlers: Dict[str, List[str]] = dict.fromkeys(HANDLERS_MAPPING, [])
+        handlers = {key: [] for key in HANDLERS_MAPPING.keys()}
         assert handlers is not None
         for doc in document_ids:
             try:
-                handlers[doc[doc.rfind("/") + 1 : doc.rfind("-")]].append(doc)
+                handlers[doc[doc.rfind("/") + 1 : doc.find("-")]].append(doc)
                 # Basename for local syncs, document_id should not have slash otherwise.
             except KeyError:
                 error_msg = f"No handler defined for document identifier {doc}"
@@ -598,8 +600,9 @@ def sync_documents(
 
     stats = dict.fromkeys(HANDLERS_MAPPING, (0, 0, 0, 0))
     for handler, documents in (handlers or static_handlers).items():
-        stats[handler] = HANDLERS_MAPPING[handler](
-            documents, force=force, graceful=graceful, graph=graph, is_local=is_local
-        )
+        if documents != []:
+            stats[handler] = HANDLERS_MAPPING[handler](
+                documents, force=force, graceful=graceful, graph=graph, is_local=is_local
+            )
 
     return stats


thoth/storages/sync.py
```diff
@@ -598,8 +598,9 @@ def sync_documents(
 
     stats = dict.fromkeys(HANDLERS_MAPPING, (0, 0, 0, 0))
     for handler, documents in (handlers or static_handlers).items():
-        stats[handler] = HANDLERS_MAPPING[handler](
-            documents, force=force, graceful=graceful, graph=graph, is_local=is_local
-        )
+        if documents is None or documents:
```
codificat, line 601: How could `documents` be `None` if it is initialized above to `[]`?

And even in that case, shouldn't that be:

```suggestion
        if documents:
```

codificat (commented):
Only one comment/question/clarification, otherwise lgtm (but I'll wait for an answer).

In any case,
/approve

thoth/storages/sync.py
harshad16: The function sync_documents default for var `document_ids` is None, if no value is provided for `document_ids` https://github.com/thoth-station/storages/blob/f8d5406d45811ce38bd703f3726614cd571a7553/thoth/storages/sync.py#L571
so , when its `document_ids` is None, it would fall into the else condition, where static_handlers are set for value None
https://github.com/thoth-station/storages/blob/f8d5406d45811ce38bd703f3726614cd571a7553/thoth/storages/sync.py#L596
in this case, when for loop over static_handler
https://github.com/thoth-station/storages/blob/f8d5406d45811ce38bd703f3726614cd571a7553/thoth/storages/sync.py#L600
the documents would be `None`

codificat: I see, thanks.

But: do we want to try to call the handler if there are no documents is `None`?

harshad16: There are 3 values in play here: `actual document id, None, []`
- When the actual document id is present, the idea is to call only that handle(for example `sync_solver_document`), so that would only get synced, and other sync functions are not called.
- When None values, the idea is to execute all the sync functions, so they can sync the all document present ceph.

- when [] is value, it just states, one of the handlers is having document_id. so executing it not all the functions should be executed. Only the handler with document_id. 

please let me know if you feel otherwise.

VannTen: I would rather have 
```suggestion
        if documents != []:
```
Which look more obvious on why we're excluding that handler.
Or maybe we could turn handlers into a defaultdict

VannTen: Although I might have had typing incompatibilites while working on that piece of code which prevented me to do just that, can't remember exactly.

goern:
shall we merge this?!

thoth/storages/sync.py
harshad16: Made the change suggested.
Ready for review again.

KPostOffice: +1 on the defaultdict because then there is no need to check for `[]`

harshad16: i m unable to understand the use of defaultdict in here.
defaultdict would still require a value.
and still would have to check that right?

VannTen: The point of `defaultdict` would be that we don't need to check for `[]` in `handlers`, because `handlers` member are only populated with a handler key the first time we add a document to it.

However, the code is fine by me in its current form and better is the enemy of good.

harshad16:
thanks all for the review
Merging for the release.

VannTen:
https://sentry.io/organizations/red-hat-it/issues/3881412534/?alert_rule_id=529086&alert_timestamp=1674010617326&alert_type=email&environment=ocp4-stage&project=1298083

Looks like we still have problems on that, and can't rely on clever string cutting...

VannTen:
Or maybe we can with enough shoehorning ! :rofl:

PR #2758, opened 2023-01-18
Fix docs sync, take 2

VannTen:
followup to #2751
it's an ugly hack but it should work'


codificat:
> it's an ugly hack but it should work'

Agreed.
/lgtm
/approve

VannTen:
Fix pre-commit failure which slipped through

